PR #29, opened 2019-10-07
One off update of all Javascript dependencies.

mozbhearsum:
To get us to a more recent stable where Dependabot can more reasonably keep things up to date. Most of these seem safe, although the big Taskcluster client change could be problematic. This should be testing in `testing` before even pushing to `staging`.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -6,12 +6,12 @@
   "author": "Rail Aliiev <rail@mozilla.com>",
   "license": "MPLv2",
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


mozbhearsum:
After some testing locally and in the `testing` environment it looks like all of the major package updates (React, Auth0, Bootstrap, Taskcluster) break one thing or another in Ship It. I don't have the expertise to resolve them quickly, so I'll have to leave it for someone more familiar with the code. I'll leave this open for now, as a reminder that it needs to get done.

PR #47, opened 2019-10-18
Bug 1589437 - Update Shipit's RC detection to account for Fennec ESR

JohanLorenzo:


JohanLorenzo:
I'll land the patch on Monday in order to avoid any issues over the weekend.

PR #54, opened 2019-10-22
Update Firefox Nightly version to 72.0a1

pascalchevrel:
We already shipped nightly 72 builds so this should be updated


JohanLorenzo:
Dupe of #49

PR #78, opened 2019-11-06
Explicitly use TASKCLUSTER_ROOT_URL

rail:
This is in staging now, but I cannot fully verify it until we have CI builds published there. I'll be doing some tests as soon as we get those. The API calls look sane to me so far in local dev.

Consider this as a feedback review.

* Remove some unused stuff
* Drop unused taskclusterProxy support and simplify `get_options()`
* Set Taskcluster root URL via CLI or environment
* `black` to print the diff on failures
* Use configuration files for setting up environment for `frontend/scripts/deploy` instead of Taskcluster secrets.
* New `taskcluster-web-client` requires setting `url`
* Stop publishing image.tar artifact


frontend/package.json
```diff
@@ -37,7 +37,8 @@
     "react-interval": "^2.0.2",
     "react-router-bootstrap": "^0.24.4",
     "react-router-dom": "^4.2.2",
-    "taskcluster-client-web": "^5.0.1",
+    "taskcluster-client-web": "^8.0.0",
```
mozbhearsum: Hooray for upgrading!

api/src/backend_common/auth0.py
```diff
@@ -59,6 +59,7 @@ def auth0_login():
     API Endpoint: Build Url to login on Auth0 server
     """
     params = {
+        # TODO: Revisit this value
```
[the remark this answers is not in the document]
rail: I don't want to touch this before the migration to reduce the amount of moving parts. Maybe I should just convert this into a bug.

frontend/scripts/env.production
```diff
@@ -0,0 +1,2 @@
+SHIPIT_API_URL=https://shipit-api.mozilla-releng.net
```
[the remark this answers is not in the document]
rail: I think, using `.taskcluster.yml` is a good idea. Thank you! I'll update the PR.

rail:
I verified this in staging and landed some minor fixes. This is ready to land on Friday! \o/

.taskcluster.yml
```diff
@@ -23,7 +23,20 @@ tasks:
         $if: 'release_channel == "production"'
         then: 'relengstatic-prod-shipitfrontend-static-website'
         else: 'relengstatic-${release_channel}-shipitfrontend-static-website'
-
+      taskcluster_root_url:
+        $if: 'release_channel == "production"'
+        then: 'https://taskcluster.net'
```
mozbhearsum: It's still early so I may be missing something...but this isn't is going away in a few days, isn't it?

I'm also not sure why we'd be using different Taskcluster instances between production and other environments.

rail: You are right, we can switch to the new cluster here and it'll be only applied when we merge to production. The instance types are already changed, just need to merge from master

rail: You are right, we can switch to the new cluster here and it'll be only applied when we merge to production. The instance types are already changed, just need to merge from master

rail:
Thank you!

PR #85, opened 2019-11-18
Add ability to skip phases

### LINKED ISSUE #84: support triggering a later phase only and not require earlier taskgroups/phases
I believe the snowman/action-task logic supports this already but shipit itself may need to be told that this is something valid.

rail:
Fixes #84

### DIFF
--- a/api/src/shipit_api/api.py
+++ b/api/src/shipit_api/api.py
@@ -169,6 +169,12 @@ def schedule_phase(name, phase):
     phase.task_id = result["status"]["taskId"]
 
     phase.submitted = True
+    # Make sure that if one of the previous phases are either submitted or
+    # skipped
+    for ph in phase.release.phases:
+        if ph.name == phase.name:
+            break
+        ph.submitted = True
     phase.completed_by = current_user.get_id()
     completed = datetime.datetime.utcnow()
     phase.completed = completed
--- a/api/src/shipit_api/api.py
+++ b/api/src/shipit_api/api.py
@@ -176,15 +176,19 @@ def schedule_phase(name, phase):
     phase.task_id = result["status"]["taskId"]
 
     phase.submitted = True
-    # Make sure that if one of the previous phases are either submitted or
-    # skipped
+    completed = datetime.datetime.utcnow()
+    phase.completed_by = current_user.get_id()
+    phase.completed = completed
+    # If the previous phases are not submitted, mark them as submitted and they
+    # will be calculated as skipped because they don't have taskId associated
     for ph in phase.release.phases:
         if ph.name == phase.name:
             break
-        ph.submitted = True
-    phase.completed_by = current_user.get_id()
-    completed = datetime.datetime.utcnow()
-    phase.completed = completed
+        if not phase.submitted:
+            ph.submitted = True
+            ph.completed_by = current_user.get_id()
+            ph.completed = completed
+
     if all([ph.submitted for ph in phase.release.phases]):
         phase.release.status = "shipped"
         phase.release.completed = completed
@@ -213,7 +217,7 @@ def abandon_release(name):
             abort(401, f"required permission: {required_permission}, user permissions: {user_permissions}")
 
         # Cancel all submitted task groups first
-        for phase in filter(lambda x: x.submitted, release.phases):
+        for phase in filter(lambda x: x.submitted and not x.skipped, release.phases):
             try:
                 actions = fetch_artifact(phase.task_id, "public/actions.json")
                 parameters = fetch_artifact(phase.task_id, "public/parameters.yml")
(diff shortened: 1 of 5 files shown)


api/src/shipit_api/api.py
```diff
@@ -169,6 +169,12 @@ def schedule_phase(name, phase):
     phase.task_id = result["status"]["taskId"]
 
     phase.submitted = True
+    # Make sure that if one of the previous phases are either submitted or
+    # skipped
```
mozbhearsum: I'm having trouble understanding this comment. "if one of" makes me think there should be some additional clause at the end. Can you rephrase this?

api/src/shipit_api/api.py
```diff
@@ -169,6 +169,12 @@ def schedule_phase(name, phase):
...
+    for ph in phase.release.phases:
+        if ph.name == phase.name:
+            break
+        ph.submitted = True
```
mozbhearsum: Just to make sure I understand this block correctly: It is marking all earlier phases as submitted -- right? It's going through the phases (which are ordered by id, if I understand the model correctly), and marking everything prior to the phase being submitted as "submitted".

api/src/shipit_api/models.py
```diff
@@ -90,11 +90,18 @@ def rendered_hook_payload(self, extra_context={}):
...
+        # in case we skip a phase, the task ID is not defined, we want to
+        # filter it out
+        resolved_previous_graph_ids = filter(None, resolved_previous_graph_ids)
+        context["input"]["previous_graph_ids"] = list(resolved_previous_graph_ids)
         if extra_context:
             context.update(extra_context)
         return render_action_hook(self.task_json["hook_payload"], context)
 
+    @property
+    def skipped(self):
```
mozbhearsum: :+1: to having this be a calculated property, rather than a new column.

api/src/shipit_api/api.py
mozbhearsum: Also, do `completed_by` or `completed` need to get set for the skipped phases, too?

api/src/shipit_api/api.py
```diff
@@ -169,6 +169,12 @@ def schedule_phase(name, phase):
...
     phase.completed_by = current_user.get_id()
     completed = datetime.datetime.utcnow()
     phase.completed = completed
```
mozbhearsum: I noticed that `abandon_release` has some logic that loops over phases and tries to fetch artifacts for them. Is that going to work for skipped phases? I see that it catches `ArtifactNotFound`, but I'm not sure if that's what will be thrown for taskIds that don't exist.

frontend/src/views/ListReleases/index.js
```diff
@@ -522,15 +534,17 @@ class TaskLabel extends React.PureComponent {
   };
 
   render() {
-    const { status, name, taskGroupUrl } = this.props;
-    if (status === 'blocked') {
+    const {
+      status, name, taskGroupUrl, tasksInProgress, allowPhaseSkipping,
+    } = this.props;
+    if ((status === 'blocked' && !allowPhaseSkipping) || status === 'skipped' || tasksInProgress) {
```
mozbhearsum: I'm having trouble evaluating this because I don't understand what `blocked` means as a phase status. Can you explain?

rail:
> While looking to see if other code needs to be updated I noticed that `shipit_api.config.SIGNOFFS` is empty, and doesn't appear to be updated by anything. Is that intentional?

Yeah. The feature is implemented, but we don't use it for anything. So we left it like that to have an example of a signoff config.

api/src/shipit_api/api.py
rail: How about "If the previous phases are not submitted, mark them as submitted and they will be calculated as skipped because they don't have taskId associated"

api/src/shipit_api/api.py
rail: Correct. Probably I should add another `if` so I don't touch actually submitted phases and for the rest add a timestamp and the submitter as you suggested.

api/src/shipit_api/models.py
rail: Heh, I actually took another take on this last week to see how it looks with a column holding this value. The DB migration part was a bit painful tbh. I'll use the current approach.

api/src/shipit_api/api.py
rail: :+1:

api/src/shipit_api/api.py
rail: Good point. It will work, but better to add the skipped ones to the filter, so we don't even try to cancel them.

frontend/src/views/ListReleases/index.js
rail: In this context "blocked" means that you cannot submit it yet. The previous phase has to be submitted and the corresponding action task has to be done. If you submit too early, the action task may not find all the needed tasks from the previous graph. I rearranged some stuff here, hope it'll be a bit clearer.

api/src/shipit_api/models.py
mozbhearsum: Regardless I think a calculated property makes sense. Otherwise we could end up in a state where taskId is not null, and skipped is true. Best to make it impossible to get into unknown states :)

PR #86, opened 2019-11-18
Add table/endpoints to enable/disable products/branches

mozbhearsum:
Very simple CRUD. Fully testable with the swagger UI.

### DIFF
--- a/api/src/shipit_api/api.py
+++ b/api/src/shipit_api/api.py
@@ -3,6 +3,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from collections import defaultdict
 import datetime
 
 from flask import abort, current_app, jsonify
@@ -15,7 +16,7 @@ from backend_common.auth import auth
 from cli_common.log import get_logger
 from cli_common.taskcluster import get_service
 from shipit_api.config import PROJECT_NAME, PULSE_ROUTE_REBUILD_PRODUCT_DETAILS, SCOPE_PREFIX
-from shipit_api.models import Phase, Release, Signoff
+from shipit_api.models import Phase, Release, Signoff, DisabledProducts
 from shipit_api.release import Product
 from shipit_api.tasks import ArtifactNotFound, UnsupportedFlavor, fetch_artifact, generate_action_hook, render_action_hook
 
@@ -323,7 +324,11 @@ def phase_signoff(name, phase, uid):
 
 
 def get_disabled_products():
-    return {}
+    session = current_app.db.session
+    ret = defaultdict(list)
+    for row in session.query(DisabledProducts).all():
+        ret[row.product].append(row.branch)
+    return ret
 
 
 def disable_product(product, branch):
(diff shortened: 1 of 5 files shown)


api/src/shipit_api/api.yml
```diff
@@ -229,6 +229,53 @@ paths:
...
+      parameters:
+      - name: product
+        in: query
+        required: true
+        type: string
+        enum:
+        - firefox
+        - fennec
+        - devedition
+        - thunderbird
```
rail: Can we limit this to firefox and devedition for now? We are not going to use this for other products at this point.

api/src/shipit_api/api.yml
```diff
@@ -229,6 +229,53 @@ paths:
...
+      - name: branch
+        in: query
+        required: true
+        type: string
```
rail: Can you add limit this to beta and try only for now? Just in case.

api/src/shipit_api/api.yml
```diff
@@ -416,3 +463,21 @@ definitions:
...
+    - branch
+    properties:
+      product:
+        type: string
+        example: firefox
+        enum:
+        - firefox
+        - fennec
+        - devedition
+        - thunderbird
```
rail: The same here, firefix and devedition only.

api/src/shipit_api/api.yml
```diff
@@ -416,3 +463,21 @@ definitions:
...
+      branch:
+        type: string
+        example: projects/maple
```
rail: The same here, limit to beta and try

api/src/shipit_api/models.py
```diff
@@ -229,3 +229,10 @@ def json(self):
             "completed": self.completed or "",
             "phases": [p.json for p in self.phases],
         }
+
+
+class DisabledProduct(db.Model):
+    __tablename__ = "shipit_api_disabled_products"
+
+    product = sa.Column(sa.String, nullable=False, primary_key=True)
```
rail: Not sure if using `primary_key` would let you add multiple branches for the same product and vice versa. Probably you want the combination to be unique.

api/src/shipit_api/api.py
```diff
@@ -320,3 +321,33 @@ def phase_signoff(name, phase, uid):
     notify_via_irc(r.product, f"{phase} of {r.product} {r.version} build{r.build_number} signed off by {who}.")
 
     return dict(signoffs=signoffs)
+
+
+def get_disabled_products():
```
rail: We also need to auth-protect these methods.

api/src/shipit_api/models.py
mozbhearsum: Setting `primary_key` on both columns turns it into a composite key. From https://docs.sqlalchemy.org/en/13/core/metadata.html:
> Multiple columns may be assigned the primary_key=True flag which denotes a multi-column primary key, known as a composite primary key.

api/src/shipit_api/api.py
mozbhearsum: Just enable/disable, or also the getter?

mozbhearsum:
The latest push should address everything, as long as we don't need to protect the list of disabled products/branches behind scopes.

api/src/shipit_api/api.py
rail: get is readonly, no need for auth

mozbhearsum:
@rail - this probably deserves another look after updating for openapi 3.0.

rail:
Ooops, sorry, I should have waited before I merged my PR. :/

PR #89, opened 2019-11-22
Automatic partials

rail:
* Partials can be automatically assigned if "auto" is passed
* Automatic partials enabled for automatic betas only. Enabling them for
other branches requires handling more edgecases, like multiple branches
for RCs and early ESR releases
* Remove ESR60

### DIFF
--- a/api/src/shipit_api/models.py
+++ b/api/src/shipit_api/models.py
@@ -167,9 +167,7 @@ class Release(db.Model):
             input_common["release_enable_emefree"] = False
 
         if self.partial_updates:
-            input_common["partial_updates"] = {}
-            for partial_version, info in self.partial_updates.items():
-                input_common["partial_updates"][partial_version] = {"buildNumber": info["buildNumber"], "locales": info["locales"]}
+            input_common["partial_updates"] = self.partial_updates
         target_action = find_action("release-promotion", self.actions)
         kind = target_action["kind"]
         if kind != "hook":
--- a/api/src/shipit_api/release.py
+++ b/api/src/shipit_api/release.py
@@ -133,10 +133,8 @@ def is_eme_free_enabled(product, version):
 
 def product_to_appname(product):
     """Convert product name to appName"""
-    # Limited to the automated beta products on purpose
     if product in [Product.FIREFOX.value, Product.DEVEDITION.value]:
         return "browser"
-    raise NotImplementedError(f"Automated releases are not implemented for {product}")
 
 
 def get_locales(repo, revision, appname):
--- a/api/src/shipit_api/release.py
+++ b/api/src/shipit_api/release.py
@@ -6,6 +6,8 @@
 import enum
 import re
 
+import requests
+
 from shipit_api.config import SUPPORTED_FLAVORS
 
 # If version has two parts with no trailing specifiers like "rc", we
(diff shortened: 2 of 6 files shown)


rail:
Evil black! :)

api/src/shipit_api/models.py
```diff
@@ -167,9 +167,7 @@ def generate_phases(self, partner_urls=None, github_token=None):
             input_common["release_enable_emefree"] = False
 
         if self.partial_updates:
-            input_common["partial_updates"] = {}
-            for partial_version, info in self.partial_updates.items():
-                input_common["partial_updates"][partial_version] = {"buildNumber": info["buildNumber"], "locales": info["locales"]}
+            input_common["partial_updates"] = self.partial_updates
```
mozbhearsum: Are you sure it's safe to remove the above block without adding it back elsewhere? AFAICT, this means that input for the non-auto case will go unmodified. Unless we're planning to change whatever is submitting `partial_updates` for the non-auto case (the frontend, I guess), I think `input_common["partial_updates"]` will end up incorrect in that case.

api/src/shipit_api/release.py
```diff
@@ -127,3 +129,19 @@ def is_eme_free_enabled(product, version):
...
             return True
     return False
+
+
+def product_to_appname(product):
+    """Convert product name to appName"""
+    # Limited to the automated beta products on purpose
+    if product in [Product.FIREFOX.value, Product.DEVEDITION.value]:
+        return "browser"
+    raise NotImplementedError(f"Automated releases are not implemented for {product}")
```
mozbhearsum: I'm not sure I understand why we're throwing errors about automated releases in this function. IMO this should be a simple pure function, and error handling/conditions happen outside of it. I can see a case for raising an exception if we don't know the `appname` for a `product` that's provided, but it definitely shouldn't be complaining about automated releases (and I'm pretty sure we don't have any known products that we don't know the app name of).

api/src/shipit_api/models.py
rail: It does nothing except converting the same structures. Not sure if it belongs here... We can just a proper object definition to `api.yml`, but I'm not sure if I know how to deal with values keyed by version...

api/src/shipit_api/release.py
rail: I didn't want to implement this for all products, so I added this exception in case we try to use for Thunderbird or something.

api/src/shipit_api/models.py
mozbhearsum: Wait. So that block was a no-op?

api/src/shipit_api/release.py
mozbhearsum: Can that condition be handled in the calling code instead of burying it here? This is a generic helper that's not specific to automated releases, so I don't think it's correct to put it here.

api/src/shipit_api/models.py
rail: I don't remember to be honest... Maybe the format was different back then, then we changed it... git blame is borken.

api/src/shipit_api/release.py
rail: easy!

api/src/shipit_api/models.py
rail: The only thing that I can think of here is that in the old approach we would catch the payload issues before we try submitting it to TC. Let me try playing with jsonschema a bit and see if I can validate it there.

rail:
Let me play with the last commit in staging, and we are should be ready to merge.

api/src/shipit_api/models.py
rail: I added swagger validation for this piece and made sure that the UI works as expected.

api/src/shipit_api/release.py
rail: Done

PR #90, opened 2019-11-23
Apply disabled product to TC clients

rail:
This will disable Tc based submissions in case we disable product using
the `/disabled-products` API endpoint.


api/src/shipit_api/api.py
```diff
@@ -73,9 +73,14 @@ def add_release(body):
         user_permissions = ", ".join(current_user.get_permissions())
         abort(401, f"required permission: {required_permission}, user permissions: {user_permissions}")
 
-    session = current_app.db.session
     product = body["product"]
     branch = body["branch"]
+
+    product_disabled = branch in get_disabled_products().get(product, [])
+    if current_user.type == "taskcluster" and product_disabled:
```
mozbhearsum: Can `taskcluster` move to `settings.py` or `config.py` to avoid hardcoding it?

PR #92, opened 2019-11-26
Use APP_CHANNEL in scopes

rail:
This adds the deployment environment name aka APP_CHANNEL to the required scopes. The Auth0 based logins don't require any changes, the permissions will be automatically populated at boot time. The TC clients' scopes should be updates by adding the new scopes before we land this PR. This will also solve the issue that the staging tc clients can poke some prod API endpoints.

### DIFF
--- a/api/src/shipit_api/config.py
+++ b/api/src/shipit_api/config.py
@@ -3,12 +3,14 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import os
 import pathlib
 import tempfile
 
 PROJECT_NAME = "shipit/api"
 APP_NAME = "shipit_api"
-SCOPE_PREFIX = f"project:releng:services/{APP_NAME}"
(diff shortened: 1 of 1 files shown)


mozbhearsum:
> staging tc clients can poke some prod API endpoints.

Um, yes please!

rail:
> > staging tc clients can poke some prod API endpoints.
> 
> Um, yes please!

Well, this is not exactly true (because they are in different clusters), but still not an ideal situation.

rail:
> Seems sensible. I assume this can't land until the appropriate scope changes are made?

Correct. I'm going to:
1. Add new scopes to the existing clients
2. Deploy this change
3. Remove the old scopes

rail:
Doh. Just realized that in production `APP_CHANNEL`  is not set in the environment, instead it comes from the secrets....

MihaiTabara:
@mozbhearsum pointed me this PR, just wanted to say +1 to adding this. Should make the scopes a bit more narrow between testing/staging/prod. 🚢

rail:
closing in favour of #96

PR #94, opened 2019-11-27
Add UI for enabling/disabling automated releases

mozbhearsum:
This PR adds UI to enable/disable automated releases. I think it's important that this UI is fairly visible and obvious, so I opted to put in a footer that shows up on all pages. I also looked at putting it in the navbar, but it seemed really weird/awkward to have nav + non-nav stuff in there. I could be convinced otherwise, though.

Here's the basic look:
image

When you click on one of the boxes, a modal dialog comes up to confirm the enable/disable action:
image

If an error occurs while trying to make a change, it shows up in the modal. If no error occurs, the page reloads with the new state.
image


MihaiTabara:
Holy smokes, this is awesome! <3 Great job @mozbhearsum

PR #95, opened 2019-11-27
Add `pending` to `tasksInProgress` check

rail:


frontend/src/views/ListReleases/index.js
```diff
@@ -348,7 +348,7 @@ class TaskProgress extends React.Component {
       const signoffs = await phaseSignOffs(releaseName, phase.name);
       return { ...phase, status, signoffs };
     }));
-    const tasksInProgress = phasesWithStatus.some(phase => phase.status === 'running');
+    const tasksInProgress = phasesWithStatus.some(phase => ['running', 'pending'].includes(phase.status));
```
mozbhearsum: Should we include `unscheduled` here, too?

rail: Added!

rail: Good catch.

PR #96, opened 2019-11-27
Add environment to scopes to separate staging/prod

rail:
This is take 2 of PR #92. This will work regardless of whether APP_CHANNEL can be determined compile time (os.environ) or run time (secrets).


mozbhearsum:
@rail and I talked about this and decided that it would be better to punt on it until after we get rid of the usage of the Taskcluster secrets service that's currently used to configure the deployed environments. The intention is to move all of that to CloudOps' traditional way of configuring secrets.

mozbhearsum:
> @rail and I talked about this and decided that it would be better to punt on it until after we get rid of the usage of the Taskcluster secrets service that's currently used to configure the deployed environments. The intention is to move all of that to CloudOps' traditional way of configuring secrets.

Whoops, this was meant for #92

PR #108, opened 2019-12-03
Remove Sylvestre from Ship-it signoff

MihaiTabara:


rail:
black's not happy, once you touch the line, you have to follow black's not that sane suggestions :/

MihaiTabara:
I'll let this ride with the next deployment since it's not urgent.

PR #124, opened 2019-12-13
Bug 1568466 - part 1: Enable local instance to schedule Fenix release promotion

JohanLorenzo:


api/src/shipit_api/api_public.yml
```diff
@@ -153,9 +153,10 @@ components:
           type: string
```
MihaiTabara: Random nit as I saw this PR flying by: good job on remembering to run https://github.com/mozilla-releng/shipit/blob/master/api/scripts/generate_public_api_schema :) I *didn't* last week and had to fight over some auth issues in papertrail :)

api/src/shipit_api/api.py
```diff
@@ -87,15 +89,16 @@ def add_release(body):
...
         branch=branch,
-        revision=body["revision"],
         build_number=body["build_number"],
-        release_eta=body.get("release_eta"),
-        status="scheduled",
         partial_updates=partial_updates,
         product_key=body.get("product_key"),
+        product=product,
+        release_eta=body.get("release_eta"),
+        repo_url=body["repo_url"],
```
rail: Hmmm, I think this may prevent automated betas to happen, unless we patch shipitscript to pass `repo_url`

frontend/src/components/vcs.js
```diff
@@ -0,0 +1,95 @@
+import { SHIPIT_API_URL } from '../config';
```
MihaiTabara: So nice to see support for both VCS systems baked in single functions, nice work!

frontend/src/views/NewRelease/index.js
```diff
@@ -69,10 +69,9 @@ export default class NewRelease extends React.Component {
       buildNumber: 0,
       partialVersions: [],
     });
-    const pushes = await getPushes(branch.repo);
-    const suggestedRevisions = Object.values(pushes.pushes).map(push =>
-      ({ ...push.changesets[0], date: new Date(push.date * 1000) })).reverse().filter(push =>
-      push.desc.indexOf('DONTBUILD') === -1);
+    const { accessToken } = this.context.authController.getUserSession();
```
rail: Hmm, I don't think you can use Auth0 access token for github... IIRC there is a special API end point to list github commits, so we don't require any auth for end users.

This applies to all github functions in this PR.

api/src/shipit_api/api_public.yml
JohanLorenzo: Hahahaha, I actually didn't know 😅 The tests actually failed!

frontend/src/views/NewRelease/index.js
JohanLorenzo: Discussed over Slack. Good news, this code talks to shipit's API :smiley:

api/src/shipit_api/api.py
JohanLorenzo: Done!

frontend/src/views/NewRelease/index.js
rail: > Discussed over Slack. Good news, this code talks to shipit's API smiley

Yeah, my bad, misread the patch.

JohanLorenzo:
Thanks for looking into this and for providing the patch! Let's not add fenix to product-details for now. In my opinion, it's a phase-3 feature https://trello.com/c/ygjIzkPt/24-show-fenix-releases-on-product-details. I imported your patch in the PR. Please let me know what you think @rail :)

PR #125, opened 2019-12-13
Fix error handling

rail:
The implementation of the error handling function didn't properly
handle connexion.problem type, which is quite different from
werkzeug.exceptions.HTTPException. It doesn't have code attribute,
what made the handler fall back to HTTP 500.


rail:
> Ship it needs tests :)

YES! :)

PR #126, opened 2019-12-19
Add public api

rail:
This adds code to support public (aka read-only) API.

The spec is generated by `openapi_subset.py`, where you can specify what section of the original API spec need to be copied.


api/scripts/openapi_subset.py
```diff
@@ -0,0 +1,44 @@
+"""
```
mozbhearsum: Hm, I hadn't thought of doing something like this when we made the Balrog public API! For it, we have multiple specs which are combined with https://github.com/MicroarrayTecnologia/spec-synthase.

I have no major objections to this approach, but one thing that is nicer about splitting them is that you don't need to remember to regenerate the public spec. (You could also workaround this by generating the public one at runtime, instead of checking it in.)

api/scripts/openapi_subset.py
```diff
@@ -0,0 +1,44 @@
...
+]
+
+
+def main(input_file, output_file):
+
+    full_api = yaml.safe_load(open(input_file))
+
+    subset = {}
+    for section in sections:
+        source = dpath.util.get(full_api, section, separator=".")
```
mozbhearsum: Good thing none of our routes have a `.` in them :)

api/scripts/openapi_subset.py
```diff
@@ -0,0 +1,44 @@
...
+    "servers",
+    "paths./releases.get",
+    "paths./releases/{name}.get",
+    "paths./releases/{name}/{phase}.get",
+    "components.schemas.Phase",
+    "components.schemas.Release",
+    "components.schemas.Signoffs",
+    "components.schemas.Signoff",
+    "components.schemas.DisableProduct",
+]
```
mozbhearsum: Since this is a public API, I think it would be a good idea to version it, otherwise we may have difficulty making changes in the future.

api/scripts/openapi_subset.py
```diff
@@ -0,0 +1,44 @@
...
+import oyaml as yaml  # oyaml preserves the order of the original document
+
+# Sections to be copied from the original document, separated by "." (period).
+sections = [
+    "openapi",
+    "info",
+    "servers",
+    "paths./releases.get",
+    "paths./releases/{name}.get",
+    "paths./releases/{name}/{phase}.get",
```
mozbhearsum: I know this is only built for our purposes right now, but I think it would be a good idea to include `paths./disabled-products.get` here as well, so that others can easily query automated release status.

api/scripts/openapi_subset.py
```diff
@@ -0,0 +1,44 @@
...
+    "components.schemas.Phase",
+    "components.schemas.Release",
+    "components.schemas.Signoffs",
+    "components.schemas.Signoff",
+    "components.schemas.DisableProduct",
```
mozbhearsum: I wonder if we should just include all of `components` to avoid any potential bustage down the road where we include new endpoints but forget about their dependencies? Or, maybe the code below could be smarter, and automatically find dependent components?

api/tox.ini
```diff
@@ -40,12 +40,15 @@ depends = py37
...
 [testenv:check]
+whitelist_externals=diff
 skip_install = true
 commands =
     black --diff --check {toxinidir}
     isort --check -rc -df {toxinidir}
     pip-compile-multi verify
     flake8 {toxinidir}
+    python {toxinidir}/scripts/openapi_subset.py {toxinidir}/src/shipit_api/api.yml {envtmpdir}/api_public.yml
+    diff -u {toxinidir}/src/shipit_api/api_public.yml {envtmpdir}/api_public.yml
```
mozbhearsum: I _think_ this is verifying that you haven't forgotten to update the checked in public api yml (great idea!). Can you add a comment about it? It's not immediately obvious. Alternatively, this could be done as a unit test, which would make it more obvious by having a named function associated with it.

api/Dockerfile.postgres
```diff
@@ -0,0 +1,3 @@
+FROM postgres:11
```
mozbhearsum: This is unnecessary. This image supports a custom init script already:
```For example, to add an additional user and database, add the following to /docker-entrypoint-initdb.d/init-user-db.sh``` (From https://hub.docker.com/_/postgres)

api/scripts/openapi_subset.py
rail: I thought about that case. The function can also accept a list of strings, instead of a dot separated string. I find the latter easier to read tbh. Worst case we can change the format and pass a list in the future.

api/scripts/openapi_subset.py
rail: I looked at spec-syntase. IIRC it doesn't support OpenAPI 3.0.0 yet. Also I find it easier to deal with a single file, which can be edited using the online editor and which can be validated without merging multiple files. :man_shrugging:

api/scripts/openapi_subset.py
rail: I followed [the suggestion](https://opensource.zalando.com/restful-api-guidelines/#113) here. But I don't have a strong opinion tbh.

mozbhearsum: I'm good with this. TIL :)

api/scripts/openapi_subset.py
rail: Bah, it totally should be there. I forgot to add it for some reason.

api/scripts/openapi_subset.py
rail: I thought about being smarter here and use only used refs, but the code becomes a bit complicated. In case we forget to add something, the server won't start. I should just add test to verify the schemas.

api/tox.ini
rail: The unittest idea wins. It'll be much easier to get better error reporting from there.

rail:
> I wasn't able to run this locally due to the containers not coming up cleanly (see https://gist.github.com/mozbhearsum/de1a5f6ff6c11a735562384045224737). It's possible that could be fixed by my suggestions for postgres below -- I'm honestly not sure.

This is a race condition, because we run 3 instances of flask in guncorn in localdev. At least it's not a regression, we can address this in a different PR, I think. Also the read-only user shouldn't try to create tables.

api/Dockerfile.postgres
rail: Doh, mounting should help.

mozbhearsum:
> > I wasn't able to run this locally due to the containers not coming up cleanly (see https://gist.github.com/mozbhearsum/de1a5f6ff6c11a735562384045224737). It's possible that could be fixed by my suggestions for postgres below -- I'm honestly not sure.
> 
> This is a race condition, because we run 3 instances of flask in guncorn in localdev. At least it's not a regression, we can address this in a different PR, I think. Also the read-only user shouldn't try to create tables.

Are you sure it's not a regression? It's possible I was just lucky before, but trying to run your patch was the first time I hit it.

rail:
> Are you sure it's not a regression? It's possible I was just lucky before, but trying to run your patch was the first time I hit it.

Yeah, I've seen this multiple times, for shipit, treestatus, and tooltool, IIRC. They all use the same approach, 3 gunicorn workers running in parallel. It even happened in production for one of the projects in the past - the service would just restart by k8s and continue without creating the full db.

api/tox.ini
rail: Moved to unittests.

api/Dockerfile.postgres
rail: Done

api/scripts/openapi_subset.py
rail: added a validation check

api/scripts/openapi_subset.py
rail: Done

PR #131, opened 2020-01-02
Bug 1605073 - enable Fennec as disableable in Ship-it

MihaiTabara:


MihaiTabara:
This works like a charm in staging, merging to master and then pushing to production.

PR #139, opened 2020-01-09
Don't rely on response content in cli

rail:


api/src/shipit_api/cli.py
```diff
@@ -148,23 +150,13 @@ def shipit_import(api_from):
...
     headers = get_taskcluster_headers(url, "post", data, taskcluster_client_id, taskcluster_access_token)
-
     # skip ssl verification when working against development instances
     verify = not any(map(lambda x: x in base_url, ["localhost", "127.0.0.1"]))
-
     r = requests.post(url, headers=headers, verify=verify, data=data)
-
     r.raise_for_status()
-
-    if r.json() != {"result": "ok"}:
```
mozbhearsum: Based on what you're removing, it sounds like there's some cases that return 200 with a different body, but still worked correctly. Is that right?

rail: The endpoint doesn't have different responses for 200. It's either 200 or 500, so I thought that there is no reason to check the response itself. The status code is enough.

rail:
No worries, there was no rush with this.

PR #146, opened 2020-01-15
Split admin/public packages

rail:
This splits the code so we can deploy the public API without admin code. The admin app will have the public code available (we need it).

I could have separated the packages into 2 or 3 completely separate modules, but that makes building docker images harder - need either to copy stuff around before calling `docker build` or publish the packages.

One of the downsides of this approach is that the public app is not reloadable in localdev, I have to install without `-e`.

TODO:
- [x] adjust `.taskcluster.yml` so we generate a separate docker image for the public app
- [x] [Bug 1610538](https://bugzilla.mozilla.org/show_bug.cgi?id=1610538) setup dockerhub repos for the images
- [x] test in staging

@mozbhearsum consider this as a `feedback?`, I'm open to change the approach.


api/Dockerfile.public
```diff
@@ -0,0 +1,21 @@
+FROM python:3.7
```
mozbhearsum: I wonder if we can get around the need for separate Dockerfiles by using extras in setup.py. Eg: shipit[admin] and shipit[public]. We'd also need to parametrize the requirements, but there's very little difference in the Dockerfiles as best I can tell. This might require moving `settings.py` into the packages (which seems like it would be a good thing to do anyways).

api/setup.py
```diff
@@ -15,26 +16,42 @@
...
+            line = line.strip()
+            # Skip empty, comment lines, and line with hashes
+            if not line or line.startswith("#") or line.startswith("--"):
+                continue
+            if line.startswith("-r "):
+                req_file = line.split()[1]
+                requirements.extend(get_requirements(req_file))
+            else:
+                requirements.extend(line.split(";")[0].rstrip("  \\").split())
+    print(requirements)
```
mozbhearsum: You can remove the debug print :)

rail: ooooops :)

api/Dockerfile.public
rail: Do you know if we can use extras in the `packages` parameter of `setup()`? I'm less worried about the dependencies (`extras_require` iirc), it's mostly about not bundling the "privileged" code in the Docker image.

The difference between Dockerfiles is minuscule, but I'm not sure if we  can use something smart using the ancient dind docker version. :/

mozbhearsum: Hmm, right. I forgot that `extras` was just for dependencies. Nevermind that part :). It doesn't matter now that I realize that the admin is already not getting installed for public based on `APP_TYPE`. So that means that the only real differences between the Dockerfiles are:
* Dependencies (meh)
* settings.py
* FLASK_APP
* APP_SETTINGS

The latter three can be dealt with by not setting defaults, and making sure `docker-compose` and CloudOps both set them. What do you think?

rail: I'm less worried about the default env variables. We can pass them in k8s.

It's more about the way how we install the app. The admin app can use `-e`, but the public app can't, because we don't want to keep the whole repo around.

Using `-e` is handy for docker-compose, because you can reload the code live. Maybe we can use a separate Dockerfile for localdev?

mozbhearsum: That seems reasonable (it's what we do for Balrog though, so I'm biased :).

rail:
PR #154 is the final version of this PR

PR #151, opened 2020-01-31
Switch to the new dockerhub repo and docker tags

rail:
In bug 1610538 we created new repos for shipit docker images. At the
same time we can simplify the tag naming schema and follow the supported
branch names without any prefix.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -7,7 +7,7 @@ tasks:
       $if: 'tasks_for == "github-pull-request"'
       then: 'PR'
       else:
-        $if: 'tasks_for == "github-push" && event.ref in ["refs/heads/testing", "refs/heads/staging", "refs/heads/production"]'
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -140,8 +133,8 @@ tasks:
                   # TODO: update these when we're to push to the real repos
                   DOCKERHUB_EMAIL: 'release+dockerhub+services@mozilla.com'
                   DOCKERHUB_USER: 'mozillarelengservices'
-                  DOCKER_REPO: 'mozilla/release-services'
-                  DOCKER_TAG: '${docker_tag}'
+                  DOCKER_REPO: 'mozilla/releng-shipit-admin'
```
mozbhearsum: yay :)

rail:
> This looks fine, but it also will conflict with #146 as far as I can tell? I assume that's known and desired though.

I wanted to gradually switch to the new repos in order to minimize possible side effects of PR #146. I'm going to resolve conflicts and test it harder in staging.

PR #154, opened 2020-02-05
Split admin and public apps

rail:
* Remove staging and testing environments.
* Use the `dev` branch as default for development.
* Build separate docker images for admin and public apps. Use separate
  requirements files, Dockerfile, but the same setup.py.
* Update README.rst to reflect the current deployment scenario.
* Refresh dependencies.
* Move public code to a separate package in the same namespace
  (shipit_api).
* Extract some model methods into functions, so the models can be used
  by the public app without methods that modify the database.
* Move `Product` and `ProductCategory` enums to a separate module.


rail:
It was a bit hard to split this into logical commits without changing everything to pass CI.

rail:
No rush with the review, I won't be landing this until after the release is out next week.

.taskcluster.yml
```diff
@@ -22,17 +22,19 @@ tasks:
         # but the other environments do.
         $if: 'release_channel == "production"'
         then: 'relengstatic-prod-shipitfrontend-static-website'
-        else: 'relengstatic-${release_channel}-shipitfrontend-static-website'
+        else:
+          $if: 'release_channel == "dev"'
+          # TODO: switch to dev
```
mozbhearsum: Can we address this soon, please? :)

rail: Yes, I didn't want to do this at the same time. Once I get this deployed, I'm going to polish everything up!

PR #170, opened 2020-02-28
Bug 1618748: Switch to python-decouple

rail:
This should help with switching from Taskcluster secrets to SOPS secrets. It also increases visibility of variables coming outside of `settings.py`


api/settings.py
```diff
@@ -4,50 +4,60 @@
...
+# APP_CHANNEL is not defined as environment variable in GCP
+if not APP_CHANNEL:
+    # TODO: This section will be removed after we switch to SOPS secrets
+    # Until bug 1618454 is fixed, fall back to the Taskcluster secrets if the
+    # corresponding environment variable is not defined
+    taskcluster_secret = config("TASKCLUSTER_SECRET")
+    taskcluster_client_id = config("TASKCLUSTER_CLIENT_ID")
+    taskcluster_access_token = config("TASKCLUSTER_ACCESS_TOKEN")
+    secrets = get_secrets(taskcluster_secret, PROJECT_NAME, taskcluster_client_id=taskcluster_client_id, taskcluster_access_token=taskcluster_access_token)
+    config = Config(repository=secrets)
```
mozbhearsum: This is overriding the imported `config` name, which may end up being confusing. Maybe we can do something like this instead?
```
from decouple import config as dc_config
APP_CHANNEL = dc_config(...)
if APP_CHANNEL:
  config = dc_config
else:
  config = Config(...)
```

api/settings.py
```diff
@@ -4,50 +4,60 @@
...
+PULSE_USE_SSL = config("PULSE_USE_SSL", default=True, cast=bool)
+PULSE_CONNECTION_TIMEOUT = config("PULSE_CONNECTION_TIMEOUT", default=5, cast=int)
+PULSE_HOST = config("PULSE_HOST", default="pulse.mozilla.org")
+PULSE_PORT = config("PULSE_PORT", default=5671, cast=int)
+PULSE_VIRTUAL_HOST = config("PULSE_VIRTUAL_HOST", default="/")
+SENTRY_DSN = optional("SENTRY_DSN")
+CORS_ORIGINS = optional("CORS_ORIGINS")
+IRC_NOTIFICATIONS_OWNERS_PER_PRODUCT = optional("IRC_NOTIFICATIONS_OWNERS_PER_PRODUCT")
+IRC_NOTIFICATIONS_CHANNELS_PER_PRODUCT = optional("IRC_NOTIFICATIONS_CHANNELS_PER_PRODUCT")
+PRODUCT_DETAILS_GIT_REPO_URL = optional("PRODUCT_DETAILS_GIT_REPO_URL")
```
mozbhearsum: Yay, explicit!

api/settings.py
```diff
@@ -4,50 +4,60 @@
...
+AUTH_CLIENT_ID = config("AUTH_CLIENT_ID")
+AUTH_CLIENT_SECRET = config("AUTH_CLIENT_SECRET")
+SQLALCHEMY_DATABASE_URI = config("DATABASE_URL")
+SECRET_KEY = config("SECRET_KEY_BASE64", cast=base64.b64decode)
+
+# optional
+optional = functools.partial(config, default=None)
+
+DISABLE_NOTIFY = config("DISABLE_NOTIFY", default=False, cast=bool)
+GITHUB_TOKEN = optional("GITHUB_TOKEN")
```
mozbhearsum: Seeing this right next to each other makes me question the `optional` partial you have. Both of these are optional, but it's less obvious that `DISABLE_NOTIFY` is, because it's using `config`. I can't think of a good way to make this more clear without dropping the partial, but I wanted to call it out.

api/src/backend_common/log.py
```diff
@@ -43,7 +41,4 @@ def configure_logging():
 
 
 def configure_sentry(environment, sentry_dsn):
-    if environment not in SENTRY_CHANNELS:
-        raise ValueError("Initializing logging with environment `{}`. It should be one of: {}".format(environment, ", ".join(SENTRY_CHANNELS)))
-
```
mozbhearsum: What made this unnecessary?

api/settings.py
rail: I can create `optional_bool` :)

api/settings.py
rail: Good point, let me address this.

api/src/backend_common/log.py
rail: settings.py checks for `supported_channels` now instead of here.

api/settings.py
rail: I'll use `config` everywhere instead.

PR #171, opened 2020-02-28
Conditionally copy version.json

rail:
`version.json` is generated in `.taskcluster.yml`, so it's not
available for `docker-compose`. We can copy-or-ignore it.

### DIFF
--- a/api/Dockerfile.public
+++ b/api/Dockerfile.public
@@ -8,10 +8,10 @@ WORKDIR /app
 COPY requirements/ /app/requirements/
 RUN pip install -r requirements/public.txt
 
-COPY version.json /app/version.json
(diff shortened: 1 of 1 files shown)


api/Dockerfile.public
```diff
@@ -8,10 +8,10 @@ WORKDIR /app
 COPY requirements/ /app/requirements/
 RUN pip install -r requirements/public.txt
 
-COPY version.json /app/version.json
 COPY docker.d/init.sh /app/docker.d/
 COPY settings_public.py /app/settings_public.py
 COPY . ./src
+RUN cp -avf /src/version.json /app/version.json || :
```
MihaiTabara: This is a bit funky syntax but 👍 if it works out.

PR #172, opened 2020-02-28
Fix `cp` path

rail:


api/Dockerfile.public
```diff
@@ -11,7 +11,7 @@ RUN pip install -r requirements/public.txt
 COPY docker.d/init.sh /app/docker.d/
 COPY settings_public.py /app/settings_public.py
 COPY . ./src
-RUN cp -avf /src/version.json /app/version.json || :
+RUN cp -avf ./src/version.json /app/version.json || :
```
MihaiTabara: I should've caught this in review, mea culpa.

rail: No worries, it's hard to spot

PR #178, opened 2020-03-04
Bug 1619704 - Move settings.py under src directory

rail:


rail:
Sorry for the last minute changes. I thought, that it would be better to use this approach, where you just calculate the path instead of importing the settings module twice.

PR #180, opened 2020-03-05
Simplify worker code

rail:
* Use flask config everywhere, instead of using half flask half env variables
* removed unused payload values that we never set in the client

Worked fine in dev \o/


rail:
Not yet...

rail:
This actually is blocked by https://bugzilla.mozilla.org/show_bug.cgi?id=1618454

rail:
will be a part of another pr

PR #181, opened 2020-03-06
Simplify taskcluster interaction

rail:
* Remove always-true conditions
* Simplify taskcluster calls
* Use `app.config` instead of `os.environ` in many places


api/src/backend_common/__init__.py
```diff
@@ -26,9 +25,6 @@ def create_app(project_name, app_name, extensions=[], config=None, redirect_root
     app.name = project_name
     app.__extensions = extensions
 
-    if not app.config.get("TESTING") and os.environ.get("APP_SETTINGS"):
-        app.config.from_envvar("APP_SETTINGS")
```
mozbhearsum: I don't know how to verify it, but I assume this is an unused code path?

rail: Correct, `APP_SETTINGS` is gone after we moved `settings.py`

PR #187, opened 2020-03-10
Bug 1618454 - stop using Taskcluster secrets as app configs

rail:
We are ready to get rid of Taskcluster secrets \o/

I also changed `worker.py`, so it uses `settings.py` instead of environment variables and drops the command line parameters and amqp body - we never used those, they just make the code harder to read.

I tested this in dev and up to the worker and everything looks fine.

### DIFF
--- a/api/src/shipit_api/admin/worker.py
+++ b/api/src/shipit_api/admin/worker.py
@@ -4,74 +4,47 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import asyncio
-import json
 import logging
 
 import click
 import flask
-from decouple import Config
 
-import cli_common.pulse
-import shipit_api.common.config
-from backend_common.taskcluster import get_secrets
+from cli_common.pulse import create_consumer, run_consumer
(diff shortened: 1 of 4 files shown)


api/src/shipit_api/admin/worker.py
```diff
@@ -4,74 +4,47 @@
...
 
     return rebuild_product_details_async
 
 
 @click.command()
-@click.option("--git-repo-url", type=str, required=False, default=None)
-@click.option("--folder-in-repo", type=str, required=True, default="public/")
-@click.option("--channel", type=click.Choice(["master", "testing", "staging", "production"]), default=None)
-@click.option("--breakpoint-version", default=shipit_api.common.config.BREAKPOINT_VERSION, type=int)
-@click.option("--clean-working-copy", is_flag=True, default=True)
```
mozbhearsum: Just to be clear: we don't use this even though the default is True?

rail: The default is `True` everywhere down to the original function. `False` is not used anywhere (it was used only for local tests).

PR #190, opened 2020-03-11
Matrix notify

rail:
The config lives in `settings.py` on purpose - it's hard to pass those via env variables. Worked fine in dev in https://chat.mozilla.org/#/room/#releaseduty-dev:mozilla.org


api/src/shipit_api/admin/settings.py
```diff
@@ -137,3 +135,28 @@
...
+
+# fmt: off
+if APP_CHANNEL == "production":
+    MATRIX_NOTIFICATIONS_OWNERS_PER_PRODUCT = {
+        "thunderbird": ["rjl", "wsmwk"],
+        "default": ["sheriffduty", "ciduty", "releaseduty"]
+    }
+    MATRIX_NOTIFICATIONS_ROOMS_PER_PRODUCT = {
+        "thunderbird": [
+            "!tBWwNyfeKqGvkNpdDL:mozilla.org"  # #releaseduty:mozilla.org
```
mozbhearsum: Huh. It kindof stinks that we have to use such terrible strings for these.

api/src/shipit_api/admin/settings.py
```diff
@@ -137,3 +135,28 @@
 
 # append scopes with scope prefix and add admin group of users
 AUTH0_AUTH_SCOPES = {f"{SCOPE_PREFIX}/{scope}": list(set(users + GROUPS["admin"])) for scope, users in AUTH0_AUTH_SCOPES.items()}
+
+# fmt: off
+if APP_CHANNEL == "production":
```
mozbhearsum: I'm certainly not going to block on it, or even say that it's necessary. But if getting this out of `settings.py` is desirable, a config (toml, ini, json, yaml - whatever) that is deployed by CloudOps might be a good way to go.

rail: I totally agree that it'd be much clearer to move these out of `settings.py`. On the other hand I'm not sure if like https://github.com/mozilla-services/cloudops-infra/blob/master/projects/relengworker/k8s/charts/signing/templates/configmap-widevine.yaml and https://github.com/mozilla-services/cloudops-infra/blob/79d7e0083c6cc800c8ea85bbfef26cc2aae74799/projects/relengworker/k8s/charts/signing/templates/deployment.yaml#L60 :/

api/src/shipit_api/admin/settings.py
rail: yeah, they suck!

PR #193, opened 2020-03-11
Add XPI to `get_release()`

rail:
In PR #179 we unified the "update release status" API endpoint to
include XPI, but didn't update the response schema. Additionally,
`shipitscript` queries `/releases/$releasename` to verify that the
release has been properly updated.

* Move all XPI models to `shipit_api.common` (no changes)
* Allow multiple response schemas
* Fix `generate_public_api_schema` paths
* regenerate public API definition


rail:
> Are there any unit tests that we can add for these? Could be a followup.

Yes. I keep postponing them, but probably it's that time. :)

PR #198, opened 2020-03-24
Update dependencies

rail:
Closes #197


mozbhearsum:
Why aren't you letting pyup do this, out of curiosity?

rail:
> Why aren't you letting pyup do this, out of curiosity?

pyup is messing up with comments, so next time when I regenerate deps because I add/remove something the diff becomes unreadable. I haven't found a way to make pyup and pip-compile-multi generate the same output :/

PR #202, opened 2020-04-06
Bump nightly and fennec versions for next release cycle

JohanLorenzo:
Same as #183 but with new numbers!


rail:
I wonder if we care about `NEXT_RELEASE_DATE` and others. They are out of date, maybe we should remove those?

PR #205, opened 2020-04-07
Prevent footer from covering content

mozbhearsum:
There's probably a better way to apply this, but I'm not sure how.


rail:
Thank you!

PR #210, opened 2020-04-21
Update dependencies

rail:
Closes #208


api/src/shipit_api/admin/product_details.py
```diff
@@ -150,23 +150,23 @@ def create_index_listing_html(folder: pathlib.Path, items: typing.Set[pathlib.Pa
         def write(x):
             return html.write(f"{x}{os.linesep}")
 
-        write(f"<!doctype html>")
-        write(f"<html>")
+        write("<!doctype html>")
+        write("<html>")
         write(f"  <head>")
```
MihaiTabara, line 155: There are still some f-strings left here, not sure if this is intended or did you want to wipe them all.

rail: Oh, right. Found 2 more. I wish the linter did this for me.

PR #212, opened 2020-04-27
Remove some obselete ways of setting configuration.

tomprince:


rail:
Need to verify this by pushing to dev, I'm not sure if `process.env` works properly in the compiled version...

tomprince:
I did do some testing of this on dev, which is how I ended up adding the lint exception.

It turns out that it overrides the full string, and so `process.env` is not something that can be destructured, when using webpack. It ends up being defined by the [define plugin](https://webpack.js.org/plugins/define-plugin/) ([via](https://webpack.js.org/plugins/environment-plugin/)).

rail:
Thank you!

PR #213, opened 2020-04-27
Switch more things to use the pubic API

tomprince:
I needed to move some the signoff stuff to be loaded later, to deal with the fact that it isn't available in the public API. I'm not sure if it not being public is deliberate, or if it could be switched to public.

Also, getting this data currently fails with an exception, which causes the modal to not open. While this might be reasonable behavior, it would probably better to be explicit about it, rather than the accidental behavior it currently has.


rail:
Let me take this over and port for the new UI. It should be much easier.

rail:
This will be addressed in the new UI

PR #220, opened 2020-05-05
Add signoffs to public API

rail:


rail:
Let's postpone this for now.

PR #225, opened 2020-05-15
Enable XPI signoffs

rail:
* Fix permission checking in `phase_signoff()`
* Remove old relman users
* Some UI clean up (sorry)

### DIFF
--- a/api/src/shipit_api/admin/api.py
+++ b/api/src/shipit_api/admin/api.py
@@ -279,6 +279,12 @@ def phase_signoff(name, phase, body):
     if who in [s.completed_by for s in phase_obj.signoffs]:
         abort(409, f"Already signed off by {who}")
 
+    # signoff.permissions corresponds to the group in settings.py
+    groups = current_app.config.get("GROUPS", {})
+    group = groups.get(signoff.permissions, [])
+    if who not in group:
+        abort(401, f"User `{who} is not in the `{signoff.permissions}` group")
+
     signoff.completed = datetime.datetime.utcnow()
     signoff.signed = True
     signoff.completed_by = who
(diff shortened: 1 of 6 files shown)


api/src/shipit_api/admin/api.py
```diff
@@ -279,6 +279,12 @@ def phase_signoff(name, phase, body):
     if who in [s.completed_by for s in phase_obj.signoffs]:
         abort(409, f"Already signed off by {who}")
 
+    # signoff.permissions corresponds to the group in settings.py
+    groups = current_app.config.get("GROUPS", {})
+    group = groups.get(signoff.permissions, [])
+    if who not in group:
+        abort(401, f"User `{who} is not in the `{signoff.permissions}` group")
```
escapewindow, line 286: Nit: we may want to add a closing backtick after `{who}`.

api/src/shipit_api/admin/settings.py
```diff
@@ -72,8 +72,8 @@
         "sfraser@mozilla.com",
         "tprince@mozilla.com",
     ],
-    "firefox-signoff": ["rkothari@mozilla.com", "ehenry@mozilla.com", "jcristau@mozilla.com", "pchevrel@mozilla.com", "rvandermeulen@mozilla.com"],
-    "fenix-signoff": ["rkothari@mozilla.com", "ehenry@mozilla.com", "jcristau@mozilla.com", "pchevrel@mozilla.com", "rvandermeulen@mozilla.com"],
+    "firefox-signoff": ["jcristau@mozilla.com", "pchevrel@mozilla.com", "rvandermeulen@mozilla.com"],
```
escapewindow, line 75: Will this break anything around automatic betas? Does relman know they will need to sign off for Firefox releases?

api/src/shipit_api/admin/api.py
rail: bah, missed that

api/src/shipit_api/admin/settings.py
rail: This PR enables signoffs only for the XPI promote phase. Everything else is unaffected.

PR #227, opened 2020-05-21
Fix xpi build number

rail:
Found this bug while I was porting the old UI. To calculate the next
build number, we should query the shipit API passing xpi_name and
xpi_version. But instead of passing xpi_version we pass xpiRevision.

It looks fine in staging.


api/src/backend_common/db.py
```diff
@@ -29,7 +29,7 @@ def init_database(app):
         if os.path.isdir(migrations_dir):
             # Needed to init potential migrations later on
             # Use a separate alembic_version table per app
-            options = {f"version_table": "{app.import_name}_alembic_version"}
+            options = {"version_table": "{app.import_name}_alembic_version"}
```
escapewindow, line 32: do we want the second string to be an f-string, or does this get expanded later?

rail: Oh, good catch! It definitely should be an f-string.

PR #228, opened 2020-05-21
adjust signoffs per discussions

escapewindow:
This will adjust signoffs as follows:

- privileged webextensions will require 2 people from either the privileged admin group or releng to sign off.
- system addons will require 2 people from either the system admin group or releng to sign off.
- mozillaonline will require 1 person from the mozillaonline group or releng, plus 1 person from the privileged admin group or releng, to sign off.

In each case, releng is a backup for bus factor, and should defer to someone from the appropriate team first.

In the coming days/weeks, we should:
- potentially split `internal` privileged addons from `studies` privileged addons, to allow for stricter signoffs on the latter;
- potentially add `qa` signoff groups for `system` and/or `studies` addons;
- make sure MozillaOnline understands the ramifications of these changes;
- document this, including how and when to sign off;
- adjust the [email notifications](https://github.com/mozilla-extensions/xpi-manifest/issues/17) so the appropriate people are notified when these releases happen.

### DIFF
--- a/api/src/shipit_api/admin/settings.py
+++ b/api/src/shipit_api/admin/settings.py
@@ -71,22 +71,9 @@ ADMIN_GROUP = [
     "sfraser@mozilla.com",
     "tprince@mozilla.com",
 ]
-XPI_PRIVILEGED_ADMIN_GROUP = [
-    "rdalal@mozilla.com",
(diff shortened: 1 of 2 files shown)


escapewindow:
@rehandalal @jvehent FYI.

PR #232, opened 2020-05-27
Make XPI messages less noisy

rail:


escapewindow:
aki  13:31
hm, yeah. i like the explicit message on non-private xpis. maybe we should hold off on this til someone complains
13:31
thank you though

PR #238, opened 2020-06-02
New UI

rail:
This is an "upgrade" of the UI. :) Mostly it's a switch from Bootstrap to Material UI.

There is no significant change to the logic, except that all components have been moved from class+setState to functions+hooks (the new react approach).

Also all the HTTP operations are switched from `fetch` to `axios`, using the public API when it's possible.

I'm happy to address any comments and describe the code (by updating the comments or in person).


mozbhearsum:
> Also all the HTTP operations are switched from fetch to axios, using the public API when it's possible.

I'm curious what the reason behind this is. Are there use cases that the UI can satisfy without the admin API?

frontend/.neutrinorc.js
```diff
@@ -1,85 +1,82 @@
...
-              { from: '__heartbeat__', to: 'views/ok.html' },
-              { from: '__lbheartbeat__', to: 'views/ok.html' },
-              { from: '__version__', to: 'views/version.json' },
-            ],
-          },
+        headers: {
+          'Content-Security-Policy':
+            `connect-src ${connectSrc}; ` +
+            "default-src https://auth.mozilla.auth0.com 'none'; " +
+            "script-src 'self' 'unsafe-eval' 'unsafe-inline' https://cdn.auth0.com; " +
```
mozbhearsum, line 43: How hard is it to get rid of the `unsafe` here, and for `style-src`? If you can't get rid of them, please document why they are needed.

frontend/package.json
```diff
@@ -1,44 +1,49 @@
...
-    "font-awesome": "^4.7.0",
-    "history": "^4.7.2",
-    "json-e": "^2.3.5",
-    "mitt": "^1.1.2",
-    "moment": "^2.22.2",
+    "@date-io/date-fns": "1.x",
+    "@hot-loader/react-dom": "^16.13.0",
+    "@material-ui/core": "^4.9.10",
+    "@material-ui/icons": "^4.9.1",
+    "@material-ui/lab": "^4.0.0-alpha.49",
```
mozbhearsum, line 32: Some of these are already out of date :)

frontend/src/App.jsx
```diff
@@ -0,0 +1,133 @@
...
+import { AuthContext } from './utils/AuthContext';
+import { USER_SESSION } from './utils/constants';
+import theme from './theme';
+import Main from './Main';
+import config, { SHIPIT_API_URL, SHIPIT_PUBLIC_API_URL } from './config';
+
+axios.interceptors.request.use(config => {
+  const result = config;
+
+  if (!config.url.startsWith('http')) {
```
mozbhearsum, line 16: How does this interact with the requests to `hg.mozilla.org` and other non-backend servers?

frontend/src/components/PhaseProgress/index.jsx
```diff
@@ -0,0 +1,238 @@
...
+    }
+  };
+
+  const scheduleOrSignoff = async () => {
+    if (phase.signoffs && phase.signoffs.length > 0) {
+      const result = await phaseSignOffAction(
+        release.name,
+        phase.name,
+        selectedSignoffUID,
+        xpi ? '/xpi/signoff' : undefined
```
mozbhearsum, line 56: It's a bit weird to have this either set to a path or undefined. Given that `phaseSignOff` hardcodes the path for non-XPI releases, perhaps this should be a flag instead of a path? It wouldn't work if you give it some random other path anyways :). Then you could just pass along `xpi`.

frontend/src/components/PhaseProgress/index.jsx
```diff
@@ -0,0 +1,238 @@
...
+      );
+
+      if (!result.error) {
+        handleClose({ refresh: true });
+      }
+    } else {
+      const result = await schedulePhaseAction(
+        release.name,
+        phase.name,
+        xpi ? '/xpi/releases' : undefined
```
mozbhearsum, line 66: Same thing here.

frontend/src/components/PhaseProgress/index.jsx
```diff
@@ -0,0 +1,238 @@
...
+      );
+
+      if (!result.error) {
+        handleClose({ refresh: true });
+      }
+    }
+  };
+
+  const phasePrettyName = name => {
```
mozbhearsum, line 75: This should probably be moved into a helper function somewhere under `utils`.

frontend/src/components/PhaseProgress/index.jsx
```diff
@@ -0,0 +1,238 @@
...
+              {prettyName}
+            </Link>
+          </StepLabel>
+        </Step>
+      );
+    }
+
+    const canBeScheduled =
+      !readOnly &&
+      // don't schedule anything if something is still in progress
```
mozbhearsum, line 135: Yay comments!

frontend/src/components/PhaseProgress/index.jsx
```diff
@@ -0,0 +1,238 @@
...
+      !phases.map(p => p.tcStatus).find(st => st === 'running') &&
```
mozbhearsum, line 136: Do we definitely want to not allow, eg: scheduling Ship if Promote is still running as a policy? I can't think of a technical reason why we have to disallow this.

rail: If something is running, what actually means that the decision task is still in progress, we don't want to schedule the next phases and possibly hit a race condition, when the next phase tries to find tasks from the previous graph, but they are not scheduled by the running decision task yet. Once the decision task is finished, it unblocks the next phases, and we don't need to wait for all tasks to finish.

frontend/src/App.jsx
rail: I think this deserves a better comment. This interceptor doesn't try to change full URLs (`startsWith('http')`), but only sets the base for non fully qualified URLs, like `/releases`. I hope it makes sense.

frontend/package.json
rail: Right, need to update those!

frontend/.neutrinorc.js
rail: Hmm, maybe it's related to the fact that I run React Dev Tools.... I'll double check.

frontend/src/App.jsx
mozbhearsum: Hm, OK.

frontend/src/components/PhaseProgress/index.jsx
mozbhearsum: Ah, I see. Makes sense.

rail:
> I'm curious what the reason behind this is. Are there use cases that the UI can satisfy without the admin API?

It's just to make it easier to have the release status page available without VPN.

rail:
First of all. THANK YOU VERY MUCH. :) This is a very quick and useful review!

>     * I'm not sure exactly how to fix it, but I think you need to fix the bug where the footer overlaps active Releases. As it stands, the "Cancel" button can be hidden if Releases are below the bottom of the visible area. If we can't find a way to make the footer non-floating, perhaps we can make it possible to hide/show it? Or move the "Cancel" button further up the card?

Yeah, I tried to fix it, but nothing worked properly. Maybe it's time to rethink how to represent these data. Maybe something smaller in the main navbar...

>     * I previously mentioned that it's not obvious that clicking a phase will schedule it. Perhaps you can change the icon on hover to make it more obvious? Maybe turn it into the checkmark version on hover?

I have no strong opinion about this, but I can try to make it more distinguishable. 

>     * What do you think about adding an autorefresh for Release status?

Good point! I can add a switch to enable/disable autorefresh.

>     * Trying to create a Thunderbird Release gave me a 404 on https://hg.mozilla.org/try-comm-central/raw-file/ca57c081746b99405ebbd9418bb87496efbbbda7/mail/locales/l10n-changesets.json -- the error message displayed didn't have enough information to debug (I had to use devtools).

I'll try to reproduce and fix this. Thanks!
 
>     * "New Release" still redirects me to Releases when there's an error, after I close the modal dialog.

Will double check this.

>     * "Recent Releases" should be called "Release History" (or similar) if it shows all old Releases.

It only shows last 4 releases or so per product, so it's not quite Release History...

>     * If it does show all old Releases, adding some basic filtering (even just a text based filter) would be really helpful.

There is a bug about that. :) I'm not going to address it in this PR though, too much work. :)
 
>     * You may need to look at Virtualization for `Recent Releases` - I suspect it's going to be quite slow in production.

It's not too bad, because it doesn't show all the releases we have.

>     * "Releases" & "XPI Releases" and "Recent Releases" & "Recent XPI Releases" are basically the same pages, with slightly different data. What do you think about combining them?

Combining as in a dropdown menu or having some switches on the page to switch between them? Maybe it's not a bad idea, hmmmm.

>     * Similarly, the "New Release" and "New XPI Release" navbar items might be good to combine into one "New" nav item, that drops down or clicks down into "Release" & "XPI Release". I know the actual pages cannot be combined, but it might make sense to combine the nav items.

Good idea. I left the top nav menu a bit annoying on purpose to come up with different ideas about how to make it better. :P

>     * I got this error when trying to create a Maple Beta: `Error: (psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "shipit_api_releases_name_key" DETAIL: Key (name)=(Firefox-78.0a1-build1) already exists. [SQL: INSERT INTO shipit_api_releases (name, product, version, branch, revision, build_number, release_eta, status, created, completed) VALUES (%(name)s, %(product)s, %(version)s, %(branch)s, %(revision)s, %(build_number)s, %(release_eta)s, %(status)s, %(created)s, %(completed)s) RETURNING shipit_api_releases.id] [parameters: {'name': 'Firefox-78.0a1-build1', 'product': 'firefox', 'version': '78.0a1', 'branch': 'projects/maple', 'revision': '26f86f7dd89d103ffff52506aa38b35cef3b8a5d', 'build_number': 1, 'release_eta': '2020-06-11T05:27:00.000Z', 'status': 'scheduled', 'created': datetime.datetime(2020, 6, 3, 17, 27, 39, 739704), 'completed': None}] (Background on this error at: http://sqlalche.me/e/gkpj)`

This sometimes happens in staging. Maple and Try use the same $product-$version-$buildNumber in the name, but we don't cross the branches when query for build numbers. This should never happen in production.
>     * I think the "Disable" text on the partials toggle should be removed. The UI control makes it obvious what it does, and having text beside it makes me unsure whether or not it means "click to disable" or "it is disabled". It would probably be good to add `alt` text for a11y reasons instead.

Easy!

>     * I wasn't able to find any way to play with signoffs - is there some way I can do that in stage or local dev?

XPI `promote` [requires a signoff](https://github.com/mozilla-releng/shipit/blob/e22c18d06b1abb4f5bee495e83e875496e3d4397/api/src/shipit_api/common/config.py#L496-L524), you can play with XPI releases.

frontend/src/App.jsx
rail: Another way would be having an explicit function that constructs the full URL instead of this hook. And adds the Auth headers. I have no strong opinion tbh. The current approach kind of to enforce all the tings in one place.

frontend/src/components/PhaseProgress/index.jsx
rail: Let me rethink this...

rail:
> Trying to create a Thunderbird Release gave me a 404 on https://hg.mozilla.org/try-comm-central/raw-file/ca57c081746b99405ebbd9418bb87496efbbbda7/mail/locales/l10n-changesets.json -- the error message displayed didn't have enough information to debug (I had to use devtools).

this is a very old revision, (62.0b3). I'll ignore this for now. At least in this PR.

frontend/src/App.jsx
mozbhearsum: Yeah, I don't feel strongly.

frontend/.neutrinorc.js
mozbhearsum: We don't need it for stage/prod though, right?

rail:
> This looks fine overall. One thing I did notice when playing in stage a 404 when refresh Releases on https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/BLXTXZS2Q2iixPzh5oG7MQ/status. I'm not sure if this is a real issue or not -- nothing showed up in the UI.

I'm going to land this and address it if it comes up.

> I also think the `unsafe` parts of the CSP should be removed from deployed environments -- at least, production -- if they're only to make react devtools work.

It breaks the UI for some reason with or without devtools. I'll try to dig this up.

PR #240, opened 2020-06-12
Bug 1644817 - support for ESR78 (in shipit)

Callek:
I know rail has a pending PR that this change would bitrot, I'll adjust after he lands on monday.


api/src/shipit_api/common/config.py
```diff
@@ -31,7 +31,7 @@
 # should be using the major version of the older release, while ESR_NEXT should
 # be using the major version of the release with greater version.
 CURRENT_ESR = "68"
-ESR_NEXT = ""
+ESR_NEXT = "78"
```
rail, line 34: This will enable 2 ESR releases in bedrock, we probably want to deploy this only after ESR78 is shipped.

PR #246, opened 2020-06-18
Use canTogglePartials for reals

rail:


rail:
> The patch looks good to me. I wonder if we want to make the whole grid disappear since some products (like Fennec/Fenix) don't have partials at all. Displaying the text field may be confusing in this context.

Oh, good point. For some reason I forgot to port the old behaviour.

PR #272, opened 2020-06-29
Bump Thundebird nightlies to 80.

clokep:
Thank you! 🚀 

CC @jfx2006


api/src/shipit_api/common/config.py
```diff
@@ -72,7 +72,7 @@
 
 # Thunderbird configs
 LATEST_THUNDERBIRD_ALPHA_VERSION = "54.0a2"
-LATEST_THUNDERBIRD_NIGHTLY_VERSION = "79.0a1"
+LATEST_THUNDERBIRD_NIGHTLY_VERSION = "80.0a1"
 # TODO: Need to update this every cycle
 THUNDERBIRD_RELEASE_BRANCH = "releases/comm-esr68"
```
jfx2006, line 77: This line needs updating to esr78 as well.

clokep: @jfx2006 My assumption was that this would be updated when we're ready to release 78?

rail: Let's land this as is and address this comment in a separate PR.

PR #323, opened 2020-07-20
Add "build" group for privileged XPI

### LINKED ISSUE #288: Split privileges for builders and promoters
Especially for privileged (AMO) extensions, not everyone who should be allowed to trigger a build should be allowed to promote one.

Can we use separate groups for building and promoting?

rail:
Fixes #288


api/src/shipit_api/admin/settings.py
```diff
@@ -71,8 +71,8 @@
     "sfraser@mozilla.com",
     "tprince@mozilla.com",
 ]
+XPI_PRIVILEGED_BUILD_GROUP = ["amarchesini@mozilla.com"]
```
escapewindow, line 74: Do we have any concerns adding all of the developers to the build group?

rail: I'll leave this to the admins. I have not enough info tbh.

PR #336, opened 2020-07-26
Switch CI to taskgraph

rail:
* Build docker images using the new taskgraph feature that doesn't
  require docker-in-docker
* Pin tox install
* Switch to node 14
* Use skopeo to push the images
* Use new docker auth format

Testing:
* I diffed the generated docker images against the old ones and the difference is expected (caused by `pip install -e` vs `pip install`)
* It worked in staging just fine

PR: https://firefox-ci-tc.services.mozilla.com/tasks/groups/Xp0VUEMzQIGIXwv6qwl_NQ
Push to dev: https://firefox-ci-tc.services.mozilla.com/tasks/groups/Nk5K79tZTZK8Y20A5T8NkQ

One thing that I still need to figure out is how to build the k8s images on PRs


taskcluster/docker/skopeo/push_image.sh
```diff
@@ -0,0 +1,28 @@
...
+chmod 600 /root/.docker/config.json
+
+echo "==== Getting image ==="
+wget -O /workspace/image.tar.zst $IMAGE_URL
+unzstd /workspace/image.tar.zst
+
+
+echo "=== Pushing to docker hub ==="
+DOCKER_ARCHIVE_TAG="${DOCKER_TAG}-$(date +%Y%m%d%H%M%S)"
+skopeo --authfile /root/.docker/config.json --insecure-policy copy docker-archive:/workspace/image.tar docker://$DOCKER_REPO:$DOCKER_TAG
```
tomprince, line 22: I think you shouldn't need to use `authfile`. Looking at `skopeo-copy(1)` suggests you can but the file at `$XDG_RUNTIME_DIR/auth.json` or `~/.config/config.json`. If you use the later, you can use the old config format, and so not need to have a different secret for skopeo.

You could add a `/etc/containers/policy.json` with 
```
{
    "default": [{"type": "reject"}],
    "transports": {
        "docker-archive": {
            "": [{"type": "insecureAcceptAnything"}]
        }
    }
}
```
instead of using `--insecure-policy` I think.

taskcluster/ci/config.yml
```diff
@@ -0,0 +1,46 @@
...
+            implementation: docker-worker
+            os: linux
+            worker-type: 'b-linux'
+deploy:
+    dev:
+        website-bucket: "relengstatic-staging-shipitfrontend-static-website"
+        frontend-taskcluster-root-url: "https://firefox-ci-tc.services.mozilla.com"
+        shipit-public-api-url: "https://public-dev.shipitapi.nonprod.cloudops.mozgcp.net"
+        shipit-api-url: "https://api.shipit.staging.mozilla-releng.net"
+        extra-scopes: ["secrets:get:project/releng/shipit/deploy"]
```
tomprince, line 40: I would call this `secret` or `deploy-secret` with value `project/releng/shipit/deploy`. That will require constructing the appropriate scope to use below.

taskcluster/shipit_taskgraph/transforms/docker_push.py
```diff
@@ -0,0 +1,41 @@
...
+        scopes.extend(deploy_config["extra-scopes"])
+
+        env = job["worker"].setdefault("env", {})
+        env.update({
+            "DOCKER_TAG": config.params["docker_tag"],
+            "FRONTEND_TASKCLUSTER_ROOT_URL": deploy_config["frontend-taskcluster-root-url"],
+            "RELEASE_CHANNEL": config.params["docker_tag"],
+            "SHIPIT_API_URL": deploy_config["shipit-api-url"],
+            "SHIPIT_PUBLIC_API_URL": deploy_config["shipit-public-api-url"],
+            "WEBSITE_BUCKET": deploy_config["website-bucket"],
```
tomprince, line 38: This transform should either be split-up or renamed, given that it is used for both frontend and api pushes. Probably the parameter should be renamed as well.

taskcluster/docker/skopeo/push_image.sh
```diff
@@ -0,0 +1,28 @@
...
+test $IMAGE_URL
+
+
+echo "=== Generating dockercfg ==="
+mkdir -m 700 /root/.docker
+wget -O- $DEPLOY_SECRET_URL | jq '.secret.docker.skopeo' > /root/.docker/config.json
+chmod 600 /root/.docker/config.json
+
+echo "==== Getting image ==="
+wget -O /workspace/image.tar.zst $IMAGE_URL
```
tomprince, line 16: I'd prefer to use a `fetch` for this, rather `wget`. That'll require using `run-task`, though you can set `checkout: false` on the tasks.

taskcluster/ci/push-js/kind.yml
```diff
@@ -0,0 +1,37 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
tomprince, line 1: Depending on what is done with the `docker_push` transform, this could be combined with the previous kind.

taskcluster/ci/push-image/kind.yml
```diff
@@ -0,0 +1,39 @@
...
+        docker-image: {in-tree: skopeo}
+        max-run-time: 3600
+        env:
+            IMAGE_URL: {artifact-reference: "<push-image/public/image.tar.zst>"}
+
+jobs:
+    shipit-admin:
+        description: "Push to shipit-admin repository."
+        dependencies:
+              push-image: build-docker-image-shipit-admin
```
tomprince, line 25: Once https://phabricator.services.mozilla.com/D82807 lands in `ci/taskgraph`, these'll need to be updated.

taskcluster/ci/push-image/kind.yml
```diff
@@ -0,0 +1,39 @@
...
+              tests: tests-api-python-38
```
tomprince, line 26: Do we want the frontend to depend on the api tests and vice-versa? Not sure what the compatibility guarantee between the two is.

taskcluster/docker/skopeo/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+FROM golang:1.14
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update -qq \
+    && apt-get install -y zstd jq \
+    && apt-get clean
+
+RUN GO111MODULE=on CGO_ENABLED=0 go get -tags 'exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp' github.com/containers/skopeo/cmd/skopeo@v1.1.0
```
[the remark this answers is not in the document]
rail: Oh right, we can use this feature now! Good idea.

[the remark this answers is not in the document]
rail: Done

taskcluster/docker/skopeo/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+COPY push_image.sh /usr/local/bin/
+RUN chmod a+x /usr/local/bin/push_image.sh
+CMD ["push_image.sh"]
+
+VOLUME /workspace
```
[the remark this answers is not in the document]
rail: Done

taskcluster/docker/skopeo/push_image.sh
rail: Done

taskcluster/shipit_taskgraph/transforms/docker_push.py
rail: I split the transform into 3. On to set the deployment secret env and scopes, second for s3 and third for docker. `WEBSITE_BUCKET` is `FRONTEND_BUCKET` now.

taskcluster/docker/skopeo/push_image.sh
rail: Done. I had to set `extract: false` and uncompress uing `unzstd`, because `fetch-content` also untars, but this is not what I wanted.

taskcluster/ci/push-js/kind.yml
rail: I left the kinds separated.

taskcluster/ci/push-image/kind.yml
rail: Good idea. Now all pushes depend on all tests. Safer this way.

taskcluster/ci/push-image/kind.yml
```diff
@@ -15,25 +20,33 @@ job-defaults:
...
+              extract: false
 
 jobs:
     shipit-admin:
         description: "Push to shipit-admin repository."
         dependencies:
-              push-image: build-docker-image-shipit-admin
-              tests: tests-api-python-38
+              tests-js: tests-frontend-node-14
+              tests-api: tests-api-python-38
```
tomprince, line 39: Not sure it makes sense, but it would be possible to move these to the defaults.

rail: :+1:

PR #340, opened 2020-07-27
Bump nightly version and release dates.

tomprince:
This should land after the next nighties are complete.


rail:
We can merge and deploy when the new nightlies are ready.

PR #344, opened 2020-07-29
Bug 1608874 - Remove Fennec now that the Fenix migration has started

JohanLorenzo:
Same as https://github.com/mozilla-releng/shipit/pull/156 but for release. 

I see there are plenty of references to Fennec in the code base. We may want to clean this up later. I think it's better to prevent shipping first.

### DIFF
--- a/api/src/shipit_api/admin/release.py
+++ b/api/src/shipit_api/admin/release.py
@@ -13,7 +13,6 @@ from shipit_api.common.product import Product
 # If version has two parts with no trailing specifiers like "rc", we
 # consider it a 'final' release for which we only create a _RELEASE tag.
 FINAL_RELEASE_REGEX = r"^\d+\.\d+$"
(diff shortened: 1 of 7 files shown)


api/src/shipit_api/admin/release.py
```diff
@@ -56,7 +51,7 @@ def is_esr(version):
 
 def is_rc(product, version, partial_updates):
     if not is_beta(version) and not is_esr(version):
-        if is_final_release(version) or _is_fennec_rc_release(product, version):
+        if is_final_release(version):
             # version supports rc flavor
             # now validate that the product itself supports rc flavor
             if SUPPORTED_FLAVORS.get(f"{product}_rc"):
```
JohanLorenzo, line 57: I had to change this file because of this very line. `SUPPORTED_FLAVORS` comes from `config.py`.

PR #346, opened 2020-07-29
Fix staging fenix config

rail:


JohanLorenzo:
Thanks for fixing this line I forgot!

PR #351, opened 2020-07-31
Inject version.json on push

rail:
In order to the cloudops deployment automation work, version.json should
point to the task in which the docker image digest is printed. This
happens only after the image is pushed. This means that we need to
generate the version.json file as a part of the push task.


rail:
Now we don't need to rebuild the images on every push \o/

rail:
I pushed to dev and it worked: https://firefox-ci-tc.services.mozilla.com/tasks/groups/VGBrBTTiTE6gBs4ySou2LA
Digest verification is disabled in dev, but at least I could verify that there is version.json in the image and it points to the push task, which prints the digest. Hopefully the digest format is what the cloudops automation expects.

rail:
Good point. I'll update the PR.

PR #396, opened 2020-08-20
Bug 1630809 - Use the new partner parameters for Firefox 81 and higher.

nthomas-mozilla:
... Always send partner & EME-free enable parameters.

I wanted to have separate control of enabling partner repack and partner attribution, so modified the existing boolean and added a new one. The comparison with 81 may need to be updated for current version this lands on.

Also changes from 
* possibly disabled by ship-it (eg early betas); default enabled in gecko

to 
* always sending an explicit value for the booleans; default enabled in gecko (except for respin cases)

This is to make it more obvious what is going on.


MihaiTabara (approved):
This LGTM and makes sense.

escapewindow:
I *think* this needs to land in conjunction with the patches in https://bugzilla.mozilla.org/show_bug.cgi?id=1630809 , which are currently in need of review. No reviewers are currently flagged? I can give these a look.

PR #355, opened 2020-07-31
Add "Rebuild product-details" settings item

rail:

Issue #60, opened 2019-10-22
Add missing CLI tools


rail:
In the past we had these entries in `setup.py`:

```python
    entry_points=dict(
        console_scripts=[
            "shipit-import = shipit_api.cli:shipit_import",
            "shipit-download-product-details = shipit_api.cli:download_product_details",
            "shipit-rebuild-product-details = shipit_api.cli:rebuild_product_details",
            "shipit-trigger-product-details = shipit_api.cli:trigger_product_details",
        ],
    ),
```

If I add them back, the CLI tools won't work:
```
shipit-import                                                                                                                                                                                                                                    [180/476]
Traceback (most recent call last):                                                                                                       
  File "/usr/local/bin/shipit-import", line 6, in <module>                                                                               
    from pkg_resources import load_entry_point                                                                                           
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3250, in <module>                                        
    @_call_aside                                                                                                                         
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3234, in _call_aside                                     
    f(*args, **kwargs)                                                                                                                                                                                                                                                            
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3263, in _initialize_master_working_set                  
    working_set = WorkingSet._build_master()                                                                                                                                                                                                                                      
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 583, in _build_master                                    
    ws.require(__requires__)                                                                                                                                                                                                                                                      
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 900, in require                                          
    needed = self.resolve(parse_requirements(requirements))                                                                              
  File "/usr/local/lib/python3.7/site-packages/pkg_resources/__init__.py", line 786, in resolve                                          
    raise DistributionNotFound(req, requirers)                                                                                           
pkg_resources.DistributionNotFound: The 'shipit-api' distribution was not found and is required by the application 
```

We install the app using `pip install -e`, which doesn't create enough metadata for `pkg_resources`. Removing `-e` from solves the issue, but it means that localdev doesn't hot reload the project.

mozbhearsum:
Fixed by #415

PR #415, opened 2020-08-25
Add entry points for admin console scripts

mozbhearsum:
This seems like it should work, but doesn't. We end up with errors such as:
```
root@7dc02f518db9:/app# upload_product_details 
Traceback (most recent call last):
  File "/usr/local/bin/upload_product_details", line 33, in <module>
    sys.exit(load_entry_point('shipit-api-admin', 'console_scripts', 'upload_product_details')())
  File "/usr/local/bin/upload_product_details", line 22, in importlib_load_entry_point
    for entry_point in distribution(dist_name).entry_points
  File "/usr/local/lib/python3.8/importlib/metadata.py", line 504, in distribution
    return Distribution.from_name(distribution_name)
  File "/usr/local/lib/python3.8/importlib/metadata.py", line 177, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: shipit-api-admin
```

...for reasons that aren't clear to me.

### DIFF
--- a/api/setup.py
+++ b/api/setup.py
@@ -45,6 +45,7 @@ test_requirements = ["pytest"]
 packages = find_packages("src", include=["backend_common", "cli_common"]) + find_namespace_packages("src", include=["shipit_api.common", "shipit_api.public"])
 name = "shipit_api-public"
 req_file = "public.txt"
+entry_points = None
 
 # In order to separate the public and admin apps, use `APP_TYPE` environment
 # variable as a most portable way to parameterize setup.py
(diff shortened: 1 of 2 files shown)


rail:
See also https://github.com/mozilla-releng/shipit/issues/60

mozbhearsum:
I'm going to go ahead and replace the `pip install -e` that's breaking this with a plain `pip install`. I'm not certain #355 is going to work, and I want to make sure we have the ability to do this easily ASAP. We can probably consider backing this out once we know whether or not that works, and if we need to add similar hooks for other commands.

mozbhearsum:
@rail provided a detailed way to fix this in #60

api/setup.py
```diff
@@ -45,13 +45,22 @@ def get_requirements(req_file):
...
 
 # In order to separate the public and admin apps, use `APP_TYPE` environment
 # variable as a most portable way to parameterize setup.py
 if os.environ.get("APP_TYPE") == "admin":
     packages.extend(find_namespace_packages("src", include=["shipit_api.admin"]))
     name = "shipit_api-admin"
     req_file = "base.txt"
+    entry_points = {
+        "console_scripts": [
+            "shipit_upload_product_details = shipit_api.admin.cli:upload_product_details",
```
JohanLorenzo, line 58: Nit: 

```suggestion
            "shipit_upload_product_details = shipit_api.admin.cli:download_product_details",
```

https://github.com/mozilla-releng/shipit/blob/db32cee78544d13847269c45f39ed7281e837b5a/api/src/shipit_api/admin/cli.py#L61-L65

JohanLorenzo (approved):
Thank you very much! I tested each entry point and managed to get them running (except one, but it's just a small nit)

:shipit:

api/setup.py
mozbhearsum: Whoops, good catch!

PR #443, opened 2020-09-11
Turn partner builds on earlier in the beta cycle

mozbhearsum:
If I recall correctly, the original intent was to have partner builds on for roughly the latter have of the beta cycle, which meant we had them for b8 and above. Since we moved to a shorter beta cycle we haven't adjusted this - which means we have them on only for 2 betas per cycle now. I think it probably makes sense to enable them for b5 -- this will turn them on a week earlier, giving us a bit more time to find/fix issues.

### DIFF
--- a/api/src/shipit_api/admin/release.py
+++ b/api/src/shipit_api/admin/release.py
@@ -94,7 +94,7 @@ def is_partner_enabled(product, version, min_version=60):
     major_version = int(version.split(".")[0])
     if product == "firefox" and major_version >= min_version:
         if is_beta(version):
-            if get_beta_num(version) >= 8:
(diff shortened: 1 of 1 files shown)


mozbhearsum:
releng for review, cc @rvandermeulen and/or @jcristau for approval

escapewindow (approved):
You read my mind!

I *think* this is all that's needed. I suppose we'll see come 82.0b5?

mozbhearsum:
> You read my mind!
> 
> I _think_ this is all that's needed. I suppose we'll see come 82.0b5?

i'm 99% sure, but yes, we'll see :)

jcristau:
Sounds reasonable to me.

PR #450, opened 2020-09-16
Add mark and dale to privileged xpi build group

mozbhearsum:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1664796#c12


MihaiTabara (approved):
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1664796#c11 Dale and Mike should have access.

PR #455, opened 2020-09-21
Bump version & dates.

mozbhearsum:
AFAICT the soft freeze dates have been wrong for awhile, unless they're explicitly supposed to be one version behind the other dates.

### DIFF
--- a/api/src/shipit_api/common/config.py
+++ b/api/src/shipit_api/common/config.py
@@ -46,17 +46,17 @@ ESR_BRANCH_PREFIX = "releases/mozilla-esr"
 # day).
 # We could have used the in-tree version, but there can be race conditions,
 # e.g. version bumped, but still no builds available.
-FIREFOX_NIGHTLY = "82.0a1"
+FIREFOX_NIGHTLY = "83.0a1"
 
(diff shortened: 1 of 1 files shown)


MihaiTabara (approved):
> AFAICT the soft freeze dates have been wrong for awhile, unless they're explicitly supposed to be one version behind the other dates.

Seems like last time I totally slipped this. Looking back, Tom updated them correctly in https://github.com/mozilla-releng/shipit/pull/340/files but I failed to do for last cycle https://github.com/mozilla-releng/shipit/pull/412.

We should update the docs to mention the release dates too 👍

MihaiTabara:
@mozbhearsum I wonder whether we should adjust
*  `CURRENT_ESR` to `78`
* `ESR_NEXT` to '' now that esr68 is EOL. I think this influences product-details which at its turn is consumed by `go-bouncer` tests. 
* `ESR_BRANCH_PREFIX` to point to esr78

MihaiTabara:
Tracked in https://github.com/mozilla-releng/shipit/pull/458 to separate the PR logic in case we need to backout.

PR #488, opened 2020-10-13
Update admin list for RelEng changes.

mozbhearsum:


MihaiTabara (approved):

api/src/shipit_api/admin/settings.py
```diff
@@ -59,11 +59,13 @@
...
 # XXX: scopes/groups are hardcoded for now
 ADMIN_GROUP = [
+    "ahalberstadt@mozilla.com",
     "asasaki@mozilla.com",
     "bhearsum@mozilla.com",
-    "catlee@mozilla.com",
     "jlorenzo@mozilla.com",
+    "jmaher@mozilla.com",
     "mtabara@mozilla.com",
+    "sclements@mozilla.com",
```
MihaiTabara, line 68: Are we going to add Cam and Bob later on?

mozbhearsum: I'm not sure Bob is going to end up on the list (we can always add him later). I totally forgot about Cam though - fixed that.

PR #525, opened 2020-11-13
Bug 1568466 - Enable prod instance to ship Fenix releases

JohanLorenzo:
I `git grep`'d a few things in the codebase to ensure we don't need anything else. But the backend doesn't make any assumptions what repos it's allowed to deal with. The fontend config drives it all.

### DIFF
--- a/frontend/src/configs/production.js
+++ b/frontend/src/configs/production.js
@@ -99,6 +99,21 @@ module.exports = {
       ],
       enablePartials: true,
     },
+    {
+      product: 'fenix',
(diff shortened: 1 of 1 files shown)


bhearsum (approved):

camd (approved):

JohanLorenzo:
https://github.com/mozilla-releng/shipit/pull/528 is ready to land and be deployed in production. I think we're now ready to test Fenix on the production pipeline. I'm going to merge this patch, then.

PR #537, opened 2020-11-26
Bug 1679399 - Use mozilla-version to parse and bump version numbers

JohanLorenzo:
Relands https://github.com/mozilla/release-services/pull/2262 and https://github.com/mozilla/release-services/pull/2231

PR #539, opened 2020-12-01
Bug 1679399 - part 2: Fix Fennec version parsed as Fenix ones

JohanLorenzo:
Follows #528 and #537 up.

Production is currently busted with this error: 

```
Uncaught exception:
  File "/usr/local/lib/python3.8/site-packages/aioamqp/protocol.py", line 316, in run
    await self.dispatch_frame()
  File "/usr/local/lib/python3.8/site-packages/aioamqp/protocol.py", line 272, in dispatch_frame
    await channel.dispatch_frame(frame)
  File "/usr/local/lib/python3.8/site-packages/aioamqp/channel.py", line 115, in dispatch_frame
    await methods[frame.name](frame)
  File "/usr/local/lib/python3.8/site-packages/aioamqp/channel.py", line 548, in basic_deliver
    await callback(self, body, envelope, properties)
  File "/usr/local/lib/python3.8/site-packages/shipit_api/admin/worker.py", line 26, in rebuild_product_details_async
    await rebuild(flask.current_app.db.session, app_channel, git_repo_url, folder_in_repo, breakpoint_version)
  File "/usr/local/lib/python3.8/site-packages/shipit_api/admin/product_details.py", line 1051, in rebuild
    "all.json": get_releases(
  File "/usr/local/lib/python3.8/site-packages/shipit_api/admin/product_details.py", line 340, in get_releases
    version = parse_version(product, product_with_version[len(product.value) + 1 :])
  File "/usr/local/lib/python3.8/site-packages/shipit_api/admin/release.py", line 32, in parse_version
    return VersionClass.parse(version)
  File "/usr/local/lib/python3.8/site-packages/mozilla_version/version.py", line 41, in parse
    raise PatternNotMatchedError(version_string, (cls._VALID_ENOUGH_VERSION_PATTERN,))
<class 'mozilla_version.errors.PatternNotMatchedError'>
PatternNotMatchedError('"-1.0" does not match the pattern: re.compile(\'\\n        ^(?P<major_number>\\\\d+)\\n        \\\\.(?P<minor_number>\\\\d+)\\n        \\\\.(?P<patch_number>\\\\d+)\\n        (\\n            -beta\\\\.(?P<beta_number>\\\\d+)\\n            |-rc\\\\.(?P<release_candidate, re.VERBOSE)')
```

The reason is: since [1], we've parsed some version numbers twice. It was failing silently until [2] came out. I was able to reproduce this bug locally by following this procedure: 

```sh 
docker-compose up --build
docker-compose run api bash
shipit_rebuild_product_details --database-url="postgres://shipituser:shipitpassword@db/shipitdb" --channel production
```

[1] https://github.com/mozilla-releng/shipit/pull/528/files#diff-60e8b2b117044af9d445865c8fe0eaada09f61ac1353e8ecf6779bf77616cb11R442
[2] https://github.com/mozilla-releng/shipit/pull/537/files#diff-60e8b2b117044af9d445865c8fe0eaada09f61ac1353e8ecf6779bf77616cb11R340

### DIFF
--- a/api/src/shipit_api/admin/product_details.py
+++ b/api/src/shipit_api/admin/product_details.py
@@ -337,7 +337,19 @@ def get_releases(
 
         old_releases = typing.cast(typing.Dict[str, ReleaseDetails], old_product_details[product_file].get("releases", dict()))  # noqa
         for product_with_version in old_releases:
-            version = parse_version(product, product_with_version[len(product.value) + 1 :])
+            # product_with_version looks like "Fennec-1.0". There is nothing after the version
+            if "-" not in product_with_version:
+                raise ValueError(f'Invalid product_with_version "{product_with_version}". It must contain a -')
+            product_string, version_string = product_with_version.split("-")
+
+            # XXX Both Fennec and Fenix output to mobile_android.json. We don't want to
+            # parse Fennec version number as if it were Fenix because the parser is
+            # strict. So, this skip statement below is to make sure we don't try to
+            # parse version numbers with the wrong parser.
+            if product_string.lower() != product.value.lower():
+                continue
+
+            version = parse_version(product, version_string)
             if version.major_number >= breakpoint_version:
                 continue
             details[product_with_version] = old_releases[product_with_version]


api/src/shipit_api/admin/product_details.py
```diff
@@ -337,7 +337,16 @@ def get_releases(
 
         old_releases = typing.cast(typing.Dict[str, ReleaseDetails], old_product_details[product_file].get("releases", dict()))  # noqa
         for product_with_version in old_releases:
-            version = parse_version(product, product_with_version[len(product.value) + 1 :])
+            product_string, version_string = product_with_version.split("-")
```
jmaher, line 340: can you add a comment as to what the value of `product_with_version` is.  This seems fragile.

api/src/shipit_api/admin/product_details.py
```diff
@@ -337,7 +337,16 @@ def get_releases(
...
+
+            # XXX Both Fennec and Fenix output to mobile_android.json. We don't want to
```
jmaher, line 342: why is there an XXX heading the comment?

JohanLorenzo: Oh, I use it to warn about potential problems. https://stackoverflow.com/questions/1452934/what-is-the-meaning-of-xxx-in-code-comments

I don't mind switching to another keyword.

api/src/shipit_api/admin/product_details.py
JohanLorenzo: Good idea! Done!

api/src/shipit_api/admin/product_details.py
```diff
@@ -337,7 +337,17 @@ def get_releases(
...
         old_releases = typing.cast(typing.Dict[str, ReleaseDetails], old_product_details[product_file].get("releases", dict()))  # noqa
         for product_with_version in old_releases:
-            version = parse_version(product, product_with_version[len(product.value) + 1 :])
+            # product_with_version looks like "Fennec-1.0". There is nothing after the version
+            product_string, version_string = product_with_version.split("-")
+
+            # XXX Both Fennec and Fenix output to mobile_android.json. We don't want to
+            # parse Fennec version number as if it were Fenix because the parser is
+            # strict. So, this skip statement below is to make sure we don't try to
+            # parse version numbers with the wrong parser.
```
camd, line 346: This case is handling if the last version was a Fennec version, and this new version will be a Fenix version, right?  Hence the format mismatch and the reason we can skip parsing the Fennec versions?

camd (approved):
I think this looks right, with my limited understanding.  :)

api/src/shipit_api/admin/product_details.py
JohanLorenzo: Good call. I'm now raising a more explicit error if `-` does not exist in `product_with_version`

api/src/shipit_api/admin/product_details.py
JohanLorenzo: It's handling an assumption we had when `get_releases()` was written: `mobile_android.json` only contains Fennec version numbers. This is not true anymore, and this file gets parsed twice: one with the parser that handles Fennec version numbers, then with the Fenix one. This case is taking out strings that are not meant for one parser or the other. 

Please let me know if I should provide more details.

JohanLorenzo:
I tested this patch locally. I'm going to land it as is so it doesn't block the next Firefox betas. Although, I'm happy to address concerns in a follow-up.

api/src/shipit_api/admin/product_details.py
camd: Nope, that sounds good to me!  Thanks.  :)

PR #551, opened 2020-12-09
bforehand to XPI_PRIVILEGED_BUILD_GROUP

escapewindow:


jmaher (approved):

wagnerand:
Are we surey we are editing the right group? If I understood correctly, this is about a study, so we'd add Ben to `XPI_NORMANDY_PRIVILEGED_GROUP` instead?

jrbenny35:
> Are we sure we are editing the right group? If I understood correctly, this is about a study, so we'd add Ben to `XPI_NORMANDY_PRIVILEGED_GROUP` instead?

I am working on this bug, https://bugzilla.mozilla.org/show_bug.cgi?id=1556057. We need an extension with access to the ```telemetry``` API

wagnerand:
> > Are we sure we are editing the right group? If I understood correctly, this is about a study, so we'd add Ben to `XPI_NORMANDY_PRIVILEGED_GROUP` instead?
> 
> I am working on this bug, [bugzilla.mozilla.org/show_bug.cgi?id=1556057](https://bugzilla.mozilla.org/show_bug.cgi?id=1556057). We need an extension with access to the `telemetry` API

How is the add-on being distributed? Does it only live in the tree? Is it only used in automated tests?

jrbenny35:
> > > Are we sure we are editing the right group? If I understood correctly, this is about a study, so we'd add Ben to `XPI_NORMANDY_PRIVILEGED_GROUP` instead?
> > 
> > 
> > I am working on this bug, [bugzilla.mozilla.org/show_bug.cgi?id=1556057](https://bugzilla.mozilla.org/show_bug.cgi?id=1556057). We need an extension with access to the `telemetry` API
> 
> How is the add-on being distributed? Does it only live in the tree? Is it only used in automated tests?

Yeah tree. We will probably just upload it as a part of the patch. Should only be used for automated tests

wagnerand:
Thanks for the clarification. In that case, neither of the two groups is a perfect fit, but since we don't have a group for that, I'd lean towards `XPI_NORMANDY_PRIVILEGED_GROUP` in order to not block you on the add-ons team for what is more like a system add-on. @escapewindow @mythmon @rehandalal wdyt?

Out of curiosity, is there no way to run this add-on unsigned during tests?

escapewindow:
I'm open. This sounds like excellent fodder for a "What addon-type should I use?" faq :)

jrbenny35:
> 
> Out of curiosity, is there no way to run this add-on unsigned during tests?

Nope, the ```telemetry``` API is privileged. I tried a bunch of ways but could only get it to work on nightly and by manually installing it via the about:dedugging menu

escapewindow:
> Thanks for the clarification. In that case, neither of the two groups is a perfect fit, but since we don't have a group for that, I'd lean towards `XPI_NORMANDY_PRIVILEGED_GROUP` in order to not block you on the add-ons team for what is more like a system add-on. @escapewindow @mythmon @rehandalal wdyt?

To confuse things even more, we also have `system` if it is in fact a system addon. I don't know when we would choose `system` vs `{normandy}-privileged`, but I think it's a separate signing cert.

mythmon:
I'm a bit hung up on whether this needs to be externally signed at all. I'd like to hear more about what this extension is and how it will be used. For example, we have this xpcshell test that generates an add-on during the test with access to priveleged APIs: https://searchfox.org/mozilla-central/source/browser/components/extensions/test/xpcshell/test_ext_normandyAddonStudy.js

What kind of automated test is this going to be? If it is a test of the telemetry API, I think something like the above might be suitable. If it is more like marionette testing then maybe not.

jrbenny35:
> What kind of automated test is this going to be? If it is a test of the telemetry API, I think something like the above might be suitable. If it is more like marionette testing then maybe not.

It's Marionette testing

escapewindow:
@mythmon can you resolve the "does this need to be signed" question with @jrbenny35 and report back here? (We can close this if we don't need a signature. If we do, perhaps we need to switch to `normandy-privileged`.

wagnerand:
@mixedpuppy and @Rob--W might also know of a way to do this without signing.

mixedpuppy:
I asked the question on bugzilla about what testing this provides that is not already covered in the extension api tests.  

https://bugzilla.mozilla.org/show_bug.cgi?id=1556057#c8

mixedpuppy:
from bugzilla:  If an end to end test is desired that is fine. if it runs against a live (test) telemetry server that may be a bonus. If it doesn't, I think the current tests cover installing an addon and registering dynamic probes.

jrbenny35:
@escapewindow what are my next steps here? Thanks.

escapewindow:
@mixedpuppy does this mean we should merge this PR?

PR #558, opened 2020-12-14
Bump nightly upon 84 release.

MihaiTabara:


api/src/shipit_api/common/config.py
```diff
@@ -44,17 +44,18 @@
 # day).
 # We could have used the in-tree version, but there can be race conditions,
 # e.g. version bumped, but still no builds available.
-FIREFOX_NIGHTLY = "85.0a1"
+FIREFOX_NIGHTLY = "86.0a1"
+h
```
escapewindow, line 48: errant keystroke?

escapewindow (approved):
Thanks! Nit on the `h`

jfx2006:
Please make sure to take PR #560 when this merges.

api/src/shipit_api/common/config.py
MihaiTabara: Yes, good catch, removed it. thanks!

MihaiTabara:
> Please make sure to take PR #560 when this merges.

Thanks for that PR, will do.

MihaiTabara:
Merging to `master` for now, deploying to `production` after we ship 84 today to avoid breaking Ship-it.

PR #559, opened 2020-12-14
Added a startup script to automate startup environment exports.

Fenn-CS:
This PR adds a startup scripts that reduces the amount of repetitive tasks that has to carried at each startup.


bhearsum (approved):

bhearsum:
Thank you!

PR #562, opened 2020-12-15
RELENG-147 - Enable production instance to ship android-components

JohanLorenzo:
Same as #525 but for Android-Components


escapewindow (approved):
\o/

PR #572, opened 2020-12-22
bug 1683798 - add mobile staging forks

escapewindow:
I suppose we'll see if we can keep the `staging-` prefix...


MihaiTabara (approved):
LGTM!

PR #573, opened 2020-12-22
bug 1683798 - Remove personal forks

escapewindow:
Depends on #572 . We'll probably merge that one first, make sure they work, and then remove the personal forks.


MihaiTabara (approved):

escapewindow:
I'm going to do everything in #572. Duping.

PR #591, opened 2021-01-07
support for xpi-manifest as a directory of .yml files instead of a si…

jmaher:
…ngle large .yml file


api/src/shipit_api/admin/github.py
```diff
@@ -35,6 +35,45 @@ def get_file_from_github(owner, repo, file_path, ref):
     return content["data"]["repository"]["object"]["text"]
 
 
+def get_files_from_github(owner, repo, file_path, ref):
```
escapewindow, line 38: In the future, we may be able to memoize or lru_cache if we end up hitting github too often.

escapewindow (approved):
Thanks!

PR #592, opened 2021-01-07
Fix python autoreloading without breaking console scripts

bhearsum:
@jmaher - to test this, you can:
```
docker-compose build
docker-compose up
```

..and then try changing some python code while it's running, and make sure those changes are picked up.


jmaher:
I tested this and it works great for me locally.

jmaher (approved):
tested locally and on dev instance

bhearsum:
This worked fine in the dev environment too.

PR #622, opened 2021-02-02
Show Fenix & Android Components Releases on "Recent Releases" view (fixes #620)

### LINKED ISSUE #620: Fenix builds not listed on the Recent Releases page
Go to https://shipit.mozilla-releng.net/recent
Fenix Beta is not listed while we created it

bhearsum:
This view currently assumes that Releases have "branches" defined --
which is not currently the case for these products. This is a hacky for
that. The better fix would be change that code to make that assumption,
but I don't have time to do that at the moment.

### DIFF
--- a/frontend/src/configs/dev.js
+++ b/frontend/src/configs/dev.js
@@ -81,7 +81,11 @@ module.exports = {
       product: 'fenix',
       prettyName: 'Fenix',
       appName: 'fenix',
-      branches: [],
+      branches: [
+        {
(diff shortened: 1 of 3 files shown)


bhearsum:
> Weird. Worth a try?

It makes perfect sense in my reading of the code, for what it's worth - I think it's something that was never updated when we added A-C/Fenix support. The other products assume hg, where `branches` make more sense. This fix worked in my local env.

PR #625, opened 2021-02-05
Fix branches dropdown for fenix & a-c

bhearsum:
This is a follow-up hack to my hack from #622


MihaiTabara (approved):
LGTM - TIL about `some()` in Javascript <ashamed>

PR #628, opened 2021-02-10
add an lru_cache to xpi manifest github calls

escapewindow:
Our constant lookups may be causing problems:

image

I had this lru_cache patch lying around. I pushed it to production to try to fix the bustage, but we should merge it properly.

**[EDIT]** still broken, but probably good to take anyway.

### DIFF
--- a/api/src/shipit_api/admin/github.py
+++ b/api/src/shipit_api/admin/github.py
@@ -1,4 +1,5 @@
 import json
+from functools import lru_cache
 from urllib.parse import unquote, urlparse
 
 import requests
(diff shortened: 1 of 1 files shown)


escapewindow:
Not finding anything useful in the logs. I tried starting up a local instance, but it's showing me a blank screen. I may need to wait for help in the morning.

bhearsum:
For posterity, I'm pretty sure the actual cause of the error above is https://github.com/mozilla-releng/shipit/issues/629, and will be fixed by https://github.com/mozilla-extensions/xpi-manifest/pull/102 in this instance.

bhearsum (approved):

escapewindow:
Thanks for both the review + real fix!

PR #642, opened 2021-02-22
mergeduty updates for nightly 88

MihaiTabara:
To be merged after we have the first successful nightly later on today at 22:00 UTC.
CC @jfx2006

### DIFF
--- a/api/src/shipit_api/common/config.py
+++ b/api/src/shipit_api/common/config.py
@@ -44,17 +44,17 @@ ESR_BRANCH_PREFIX = "releases/mozilla-esr"
 # day).
 # We could have used the in-tree version, but there can be race conditions,
 # e.g. version bumped, but still no builds available.
-FIREFOX_NIGHTLY = "87.0a1"
+FIREFOX_NIGHTLY = "88.0a1"
 
 # The next 6 dates are information about the current and next release
 # They must be updated at the same time as FIREFOX_NIGHTLY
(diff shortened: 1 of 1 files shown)


api/src/shipit_api/common/config.py
```diff
@@ -44,17 +44,17 @@
...
-LAST_MERGE_DATE = "2021-01-25"
-LAST_RELEASE_DATE = "2021-01-26"
-NEXT_SOFTFREEZE_DATE = "2021-02-18"
-NEXT_MERGE_DATE = "2021-02-22"
-NEXT_RELEASE_DATE = "2021-02-23"
+LAST_SOFTFREEZE_DATE = "2021-02-18"
+LAST_MERGE_DATE = "2021-02-22 "
+LAST_RELEASE_DATE = "2021-02-23"
+NEXT_SOFTFREEZE_DATE = "2021-03-18"
+NEXT_MERGE_DATE = "2021-03-22 "
```
jmaher, line 56: nit: there is a whitespace character here which shouldn't be, same for LAST_MERGE_DATE

MihaiTabara: Good catch, fixed. thanks!

jmaher (approved):

MihaiTabara:
I'll deploy in the Tuesday morning UTC, once we have PM nightlies green.

PR #649, opened 2021-03-01
support directory in manifest

escapewindow:
This may currently be breaking shipit for addons.


jmaher (approved):
thanks, nice and simple

Issue #750, opened 2021-05-20
`docker-compose up` is busted in latest docker hub images


escapewindow:
This appears to be busted in both `python:3.8.10` and `python:3.8.9`. `python:3.8.8` appears to work.
Because we have `FROM python:3.8`, I imagine that tag has been moved to point at `python:3.8.10`, which is busted.

A workaround is to change `api/Dockerfile` to `FROM python:3.8.8` .
We may want to instead re-pin the requirements with a newer `python:3.8` image.

bhearsum:
Should be fixed by #752.

PR #752, opened 2021-05-20
Issue #750 - upgrade all dependencies

sarah-clements:
Ran `pip-compile-multi --upgrade` and then verified `docker-compose build` worked (without docker file pinned to python 3.8.8).


sarah-clements:
hrm, I don't understand why that python test failed with a dependency upgrade.

bhearsum (approved):
Thanks!

PR #759, opened 2021-05-25
Fix postgres protocol

bhearsum:


sarah-clements (approved):

docker-compose.yml
```diff
@@ -21,7 +21,7 @@ services:
       - HOST=0.0.0.0
       - PORT=8015
       - APP_CHANNEL=development
-      - DATABASE_URL=postgres://shipituser:shipitpassword@db/shipitdb
+      - DATABASE_URL=postgresql://shipituser:shipitpassword@db/shipitdb
```
sarah-clements, line 24: Actually this needs to be updated on line 60 too.

sarah-clements: And in the docs, here: https://github.com/mozilla-releng/shipit#to-rebuild-product-details

PR #760, opened 2021-05-26
remove old email addresses from shipit admin

jmaher:


api/src/shipit_api/admin/settings.py
```diff
@@ -93,11 +91,10 @@
     "jkerim@mozilla.com",
     "mbanner@mozilla.com",
     "mcooper@mozilla.com",
-    "rdalal@mozilla.com",
     "ttran@mozilla.com",
 ]
 
-XPI_SYSTEM_ADMIN_GROUP = ["rdalal@mozilla.com", "mcooper@mozilla.com"]
+XPI_SYSTEM_ADMIN_GROUP = ["mcooper@mozilla.com"]
```
MihaiTabara, line 97: @kewisch I wonder if you have thoughts here - who else should we add in the long run for this?

bhearsum (approved):

Issue #426, opened 2020-08-28
show metadata about a release


escapewindow:
Which revision should be the minimum. All of the inputs going into the action task might be nice. This goes for the current releases and recent releases views.

escapewindow:
See also: #295 . Requestor, signoff users, and a comment field would be awesome.

Fenn-CS:
Hello @escapewindow @bhearsum , I would like to work on this. Can I get some more context and get a go-head?

escapewindow:
@Fenn-CS Awesome! @bhearsum is willing to mentor. First step is probably setting up your local environment, per the [readme](https://github.com/mozilla-releng/shipit/blob/main/README.rst).

escapewindow:
<img width="867" alt="current_releases_view" src="https://user-images.githubusercontent.com/826343/101831839-5f911000-3aeb-11eb-9788-bd8b645e10cf.png">

Here are some thoughts about metadata we could expose in the `Releases` and `Recent Releases` views. I'm open about what this looks like.

Fenn-CS:
Thank you @escapewindow and yes! am very happy with the mentorship from @bhearsum thank you in advance! Meanwhile my local environment is already setup but I get `Error contacting Shipit backend: Error: Network Error. Are you connected to the VPN?` on the homescreen.

escapewindow:
Filed #553.

escapewindow:
@Fenn-CS are you still working on this?

Fenn-CS:
@escapewindow Unfortunately no, there were several things mostly permissions that I needed to work that was/is unfortunately not easily available.

escapewindow:
Sorry to hear. Thanks for giving it a try!

sarah-clements:
@bhearsum could you weigh in on what Aki mentioned in [this comment](https://github.com/mozilla-releng/shipit/issues/426#issuecomment-742810686) about displaying signoffs (presumably I'd use data from the `signoff` API?) and what signoffs are remaining. Could you clarify a bit what "remaining signoffs" would refer to?

bhearsum:
> @bhearsum could you weigh in on what Aki mentioned in [this comment](https://github.com/mozilla-releng/shipit/issues/426#issuecomment-742810686) about displaying signoffs (presumably I'd use data from the `signoff` API?) and what signoffs are remaining. Could you clarify a bit what "remaining signoffs" would refer to?

[...]

PR #761, opened 2021-05-27
Issue #426 - show metadata about releases

sarah-clements:
Changes made:
* Add revision and date to cards
* Add signoff completed_by for XPI releases
* Change labels to include 'task' to make it clearer what it will be linking to

Anything else I should add or change?
Screen Shot 2021-05-26 at 4 54 35 PM

@bhearsum how do you handle linting? It doesn't look like there's a way for me to run a script to clean up the linting and formatting, ie `yarn lint --fix` (when I do this in the frontend directory, I get an error about eslint) and `yarn format` (for prettier). What do you do?


bhearsum:
> @bhearsum how do you handle linting? It doesn't look like there's a way for me to run a script to clean up the linting and formatting, ie yarn lint --fix (when I do this in the frontend directory, I get an error about eslint) and yarn format (for prettier). What do you do?

Huh, I've used this successfully in the past. It probably doesn't work if you're on the wrong node version though. I'd be curious to see the full out from your `yarn lint --fix` attempt.

bhearsum (commented):
> Anything else I should add or change?

This looks great overall! I think there's one ot of other pieces of data to add, and I have a couple of nits that you can take or leave:
* Including the role a user signed off as (`name` from URLs like https://shipit-api.mozilla-releng.net/xpi/signoff/rally-core-addon-1.3.1-build1/promote) is probably a good idea.
* I think "Signed off by" or "Approved by" would be better text (maybe it's just me, but "Signing" implies a cryptographic action to me)
* I'm wondering if it might be nicer visually to put the list of names centered underneath the checkbox, instead of paired with the Phase text. The way it is now, the Phase name will move up or down depending on how many people have signed off.

sarah-clements:
> > Anything else I should add or change?
> 
> This looks great overall! I think there's one ot of other pieces of data to add, and I have a couple of nits that you can take or leave:
> 
> * Including the role a user signed off as (`name` from URLs like https://shipit-api.mozilla-releng.net/xpi/signoff/rally-core-addon-1.3.1-build1/promote) is probably a good idea.

> * I think "Signed off by" or "Approved by" would be better text (maybe it's just me, but "Signing" implies a cryptographic action to me)

Sure, will address these two points.

> * I'm wondering if it might be nicer visually to put the list of names centered underneath the checkbox, instead of paired with the Phase text. The way it is now, the Phase name will move up or down depending on how many people have signed off.

I agree with you, but it'll require some custom CSS to override the default styles.

sarah-clements:
> > @bhearsum how do you handle linting? It doesn't look like there's a way for me to run a script to clean up the linting and formatting, ie yarn lint --fix (when I do this in the frontend directory, I get an error about eslint) and yarn format (for prettier). What do you do?
> 
> Huh, I've used this successfully in the past. It probably doesn't work if you're on the wrong node version though. I'd be curious to see the full out from your `yarn lint --fix` attempt.

heh, yup that was the problem. Needed to upgrade node via nvm instead of brew too. Got it working now!

sarah-clements:
@bhearsum I tinkered with this a bit but with the role added, there ends up being a lot more text on the screen. And if you have that for build and ship too (or maybe just ship?) I think it'll look a bit too 'busy' and will necessitate a change in card size but then further changes will probably be needed to make sure it looks good on different screen sizes. 
Screen Shot 2021-05-27 at 4 48 52 PM

Some alternative ideas:
1) Have a line of text with 'approved' perhaps with an info or question mark icon to encourage someone to mouse over to view a tooltip with approving names and roles
2) A 'show approvals' button to toggle a div that will show approval details (open/close)

Thoughts? Other ideas?

bhearsum:
I think you're right that's it too noisy in the current state. I'm happy with either of your proposed solutions, with a slight preference for the tooltip.

sarah-clements:
Added a popover, so it'll look like this on mouseover
Screen Shot 2021-05-28 at 6 09 21 PM

frontend/src/components/PhaseProgress/MouseOverPopover.jsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
```
bhearsum, line 1: This is getting into nit territory, but this component looks like it's not really specific to `PhaseProgress`, should it live elsewhere? I'm happy either way, so your call.

bhearsum (approved):

frontend/src/components/PhaseProgress/MouseOverPopover.jsx
sarah-clements: I'll move it to a `helpers` older or something like that.

PR #777, opened 2021-06-10
Revision link

sarah-clements:
As a follow-up to #426, I made the revision sha a link if the repo url exists (and can be found) per @wagnerand's request. 
Screen Shot 2021-06-10 at 1 45 10 PM

Since the repo url's aren't stored in the `Release` and `XPIRelease` tables, they can't be returned as part of the `release` API's response so I've opted for a solution that "just works". I figured this implementation would be revisited if/when we make changes to the hard coded repo's as part of the Releng for Moz shipit improvements.

I could add a test for this but I'm not seeing any UI test files, just a jest config file.

### DIFF
--- a/frontend/src/components/ReleaseProgress/index.jsx
+++ b/frontend/src/components/ReleaseProgress/index.jsx
@@ -18,6 +18,8 @@ import useAction from '../../hooks/useAction';
 import ReleaseContext from '../../utils/ReleaseContext';
 import { AuthContext } from '../../utils/AuthContext';
 import config from '../../config';
+import { repoUrlBuilder } from '../../utils/helpers';
+import { XPI_MANIFEST_REPO } from '../../utils/constants';
 
 const useStyles = makeStyles(() => ({
   cardActions: {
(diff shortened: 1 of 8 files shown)


bhearsum:
> I could add a test for this but I'm not seeing any UI test files, just a jest config file.

Yeah, that's all there is :upside_down_face:

frontend/src/components/ReleaseProgress/index.jsx
```diff
@@ -90,17 +92,55 @@ export default function ReleaseProgress({
...
   const renderReleaseTitle = (isXPI, release) => {
+    let url = null;
+    let productBranch = null;
     const trimmedRevision = release.revision.substring(0, 13);
+    const product = config.PRODUCTS.find(
+      product => product.product === release.product
+    );
+
+    if (product && product.branches) {
+      productBranch = product.branches.find(
```
bhearsum, line 103: I'm curious why you're pulling branch from the config, but not the owner + repo. Adding a constant such as `XPI_MANIFEST_REPO` outside of the configs means that this won't work properly for local dev or stage.

sarah-clements: > I'm curious why you're pulling branch from the config, but not the owner + repo. 
> Adding a constant such as `XPI_MANIFEST_REPO` outside of the configs means that this won't work properly for local dev or stage.

Oh, somehow I missed it at the bottom of the config file (and thus didn't realize it was was different for each deployment). Will change.

frontend/src/components/ReleaseProgress/index.jsx
```diff
@@ -90,17 +91,55 @@ export default function ReleaseProgress({
...
+    // non-hg or firefox projects are formatted differently in the config files
+    if (!productBranch && product && product.repositories) {
+      productBranch = product.repositories.find(
+        item => item.project === release.product
+      );
+    }
 
     if (isXPI) {
-      return trimmedRevision;
+      url = repoUrlBuilder(config.XPI_MANIFEST.repo, release.revision);
```
sarah-clements, line 116: I changed the use of 'repo' to refer to the repo url for XPI_MANIFEST in the config files (and updated its use elsewhere), so that it's consistent with the rest of the projects.

frontend/src/configs/dev.js
```diff
@@ -119,6 +119,7 @@ module.exports = {
   XPI_MANIFEST: {
     branch: 'master',
     owner: 'mozilla-releng',
-    repo: 'staging-xpi-manifest',
+    project: 'staging-xpi-manifest',
+    repo: 'https://github.com/mozilla-releng/staging-xpi-manifest',
```
bhearsum, line 123: Can we construct this when we needed rather than duplicating the owner/project? The risk now is that they get out of date, and we get weird behaviour.

sarah-clements: hrm, is the risk different for xpi_manifest versus all other projects that do the same thing with regards to project and repo?

My thinking is that it's preferable to be consistent with the use of repo and project, rather than having to make special code changes just for the generation of xpi manifest urls. But if you feel strongly about it, I can do that. :)

sarah-clements: Btw, I agree with you in general, and that's something that can be done if we move towards storing the data in a table - we can just generate all urls dynamically.

bhearsum: Fair point - we may as well be consistent with what's there (and I'm not going to ask you to change everything else...)

bhearsum (approved):

frontend/src/configs/dev.js
sarah-clements: Not yet, at least :)

PR #782, opened 2021-06-11
RELENG-487 - s/email/ldap group/ for shipit scopes

jmaher:


escapewindow:
> RELENG-302 - s/email/ldap group/ for shipit scopes

Should we rename this `RELENG-487 - ...` ?

api/src/backend_common/auth.py
```diff
@@ -171,15 +174,18 @@ def __init__(self, token, userinfo):
...
         self.userinfo = userinfo
 
-        logger.info("Init user %s", self.get_id())
+        logger.info("Init user %s", self.get_email())
 
-    def get_id(self):
+    def get_email(self):
         return self.userinfo["email"]
 
+    def get_id(self):
```
escapewindow, line 182: `get_ldap_groups` or similar may be a better name here.

api/src/backend_common/auth.py
```diff
@@ -171,15 +174,18 @@ def __init__(self, token, userinfo):
...
+        return self.userinfo["https://sso.mozilla.com/claim/groups"]
+
     def get_permissions(self):
-        user = self.get_id()
+        users_ldap = self.get_id()
```
escapewindow, line 186: maybe `user_groups` or `user_ldap_groups` ?

api/src/backend_common/auth.py
```diff
@@ -171,15 +174,18 @@ def __init__(self, token, userinfo):
...
         all_permissions = flask.current_app.config.get("AUTH0_AUTH_SCOPES", dict())
-        return [permission for permission, users in all_permissions.items() if user in users]
+        return [permission for permission, users in all_permissions.items() if set(users_ldap).intersection(set(users))]
```
escapewindow, line 188: `users` is probably a misnomer here -- we will have a mapping of `scope` to `ldap groups`. Maybe `permission_groups`?

escapewindow (changes requested):
I think this looks good, logic-wise, thank you!
a) have we tested locally and/or on shipit-stage? and
b) some of the variable names aren't as clear as they could be. I mentioned some potential replacements.

jmaher:
I tested this locally via a local shipit instance. If this looks good, I can do it on shipit staging and get more traction there.

api/src/backend_common/auth.py
```diff
@@ -44,7 +44,7 @@ class BaseUser(object):
     type = AuthType.NONE
 
     def __eq__(self, other):
-        return isinstance(other, BaseUser) and self.get_id() == other.get_id()
+        return isinstance(other, BaseUser) and self.get_ldap_groups() == other.get_ldap_groups()
```
escapewindow, line 47: Hm, I would have thought the email might have been a better comparison?

I'm curious if we have any non-Auth0 users that use shipit; if so, some of these classes may be obsolete.

api/src/shipit_api/admin/api.py
```diff
@@ -278,19 +278,20 @@ def phase_signoff(name, phase, body):
     # Prevent the same user signing off for multiple signoffs
     phase_obj = session.query(Phase).filter(Release.id == Phase.release_id).filter(Release.name == name).filter(Phase.name == phase).first_or_404()
 
-    who = current_user.get_id()
-    if who in [s.completed_by for s in phase_obj.signoffs]:
-        abort(409, f"Already signed off by {who}")
+    users_ldap = current_user.get_id()
```
escapewindow, line 281: Hm, might be a `get_ldap_groups` ?

escapewindow (commented):
Overall looks good! I'd like to see the shipit stage results :)

escapewindow:
@bhearsum You may have some opinions about this refactor; let us know if you do?

api/src/backend_common/auth.py
```diff
@@ -132,9 +132,12 @@ def __init__(self, credentials):
 
         self.credentials = credentials
 
-        logger.info("Init user %s", self.get_id())
+        logger.info("Init user %s", self.get_email())
 
-    def get_id(self):
+    def get_ldap_groups(self):
```
bhearsum, line 137: Is there a way we can rework this so that `TaskclusterUser` is returning a sensible value here? If `get_ldap_groups` isn't a thing that makes sense in the context of a `TaskclusterUser`, it doesn't belong as part of the `BaseUser` set of methods IMO.

api/src/backend_common/auth.py
```diff
@@ -132,9 +132,12 @@ def __init__(self, credentials):
...
+        return self.credentials["clientId"]
+
+    def get_email(self):
```
bhearsum, line 140: Is this really an e-mail for a `TaskclusterUser`?

api/src/shipit_api/admin/settings.py
```diff
@@ -58,53 +58,23 @@
...
-
-XPI_MOZILLAONLINE_PRIVILEGED_GROUP = ["bzhao@mozilla.com", "jxia@mozilla.com", "yliu@mozilla.com"]
-XPI_MOZILLAONLINE_PRIVILEGED_ADMIN_GROUP = ["awagner@mozilla.com", "mkaply@mozilla.com"]
+XPI_PRIVILEGED_BUILD_GROUP = ["xpi_privileged_build"]
+
+XPI_PRIVILEGED_ADMIN_GROUP = ["xpi_privileged_admin"]
+XPI_NORMANDY_PRIVILEGED_GROUP = ["xpi_normandy_admin"]
+XPI_SYSTEM_ADMIN_GROUP = ["xpi_system_admin"]
+XPI_MOZILLAONLINE_PRIVILEGED_GROUP = ["xpi_mozillaonline_build"]
+XPI_MOZILLAONLINE_PRIVILEGED_ADMIN_GROUP = ["xpi_mozillaonline_admin"]
```
bhearsum, line 70: This is so much cleaner!

A consistency nit though: in the code above, you use `ldap_groups` a lot, but here you're just using `GROUP`. We should be consistent and either always call them "ldap groups" or never.

api/src/backend_common/auth.py
```diff
@@ -106,52 +111,11 @@ class AnonymousUser(BaseUser):
     anonymous = True
     type = AuthType.ANONYMOUS
 
-    def get_id(self):
+    def get_ldap_groups(self):
         return "anonymous:"
 
-
-class TaskclusterUser(BaseUser):
```
escapewindow, line 113: That's one way to do it :)
I don't foresee us moving back to tc auth, so this probably works.

api/src/backend_common/auth.py
```diff
@@ -247,42 +214,9 @@ def wrapper(*args, **kwargs):
...
-    except Exception as e:
-        logger.info("Taskcluster auth error for payload %s: %s", payload, e)
-        return NO_AUTH
-
-    return TaskclusterUser(resp)
-
-
-def parse_header_auth0(request):
+@auth.login_manager.request_loader
+def parse_header(request):
```
escapewindow, line 218: Hm, I think this changes the `parse_header` behavior from returning `None` if you're not authorized, to returning the `NO_AUTH` object. Let's do a little testing around this?

api/src/shipit_api/admin/xpi.py
```diff
@@ -140,19 +140,20 @@ def phase_signoff(name, phase, body):
         abort(401, f"required permission: {required_permission}, user permissions: {user_permissions}")
 
     # Prevent the same user signing off for multiple signoffs
-    who = current_user.get_id()
-    if who in [s.completed_by for s in phase_obj.signoffs]:
-        abort(409, f"Already signed off by {who}")
+    users_ldap = current_user.get_ldap_groups()
+    users_email = current_user.get_email()
+    if users_ldap in [s.completed_by for s in phase_obj.signoffs]:
```
escapewindow, line 145: Hm, we may be saving the email in the signoffs, not the ldap groups, no?

api/src/shipit_api/admin/api.py
```diff
@@ -278,19 +278,20 @@ def phase_signoff(name, phase, body):
     # Prevent the same user signing off for multiple signoffs
     phase_obj = session.query(Phase).filter(Release.id == Phase.release_id).filter(Release.name == name).filter(Phase.name == phase).first_or_404()
 
-    who = current_user.get_id()
-    if who in [s.completed_by for s in phase_obj.signoffs]:
-        abort(409, f"Already signed off by {who}")
+    users_ldap = current_user.get_ldap_groups()
+    users_email = current_user.get_email()
+    if users_ldap in [s.completed_by for s in phase_obj.signoffs]:
```
escapewindow, line 283: Same here, I think we're using the email in the signoffs list, not the ldap group?

api/src/shipit_api/admin/api.py
```diff
@@ -64,7 +64,7 @@ def add_release(body):
     branch = body["branch"]
 
     product_disabled = branch in get_disabled_products().get(product, [])
-    if current_user.type == AuthType.TASKCLUSTER and product_disabled:
+    if product_disabled:
```
escapewindow, line 67: Hm, if taskcluster auth types weren't used, this block may be obsolete.

escapewindow (commented):
I think I just have some concerns about the logic around the areas I commented. Overall this is a good cleanup, thank you!

api/src/backend_common/auth.py
escapewindow: Per https://flask-login.readthedocs.io/en/latest/_modules/flask_login/login_manager.html , 

```
The function you set should take an authentication token and
return a user object, or `None` if the user does not exist.
```

api/src/shipit_api/admin/api.py
jmaher: good catch, I have cleaned this up and a few other areas where signoff id/email/completed_by/signoffs were not clearly using the email.

api/src/backend_common/auth.py
jmaher: yeah, fixing this to be None

api/src/shipit_api/admin/api.py
jmaher: this is the only thing I am not sure about, if a product is disabled, don't we want to skip this?

escapewindow: I don't know what products are marked as disabled. The fact that this was under an `AuthType.TASKCLUSTER` if block tells me we're probably not using this, so it might be harmless.

escapewindow (approved):

PR #809, opened 2021-07-13
Notify additional people when releases change status

### LINKED ISSUE #439: notify people involved in a release
Requestor + signoff folk should be notified when a release changes status.

sarah-clements:
Resolves #439. I've tested this locally with my [xpi-manifest pr](https://github.com/mozilla-extensions/xpi-manifest/pull/120) pushed to staging-xpi-manifest and after creating an xpi release against staging-public, I saw emails (started and on completion) when a release was promoted to the next stage. However, since I was the scheduler and sign-off person for my tests, its an incomplete test so we'll want to do more of it on staging. 

Ben, it'd be good to have some tests for this. Where would you prefer I add them?

### DIFF
--- a/api/src/shipit_api/admin/xpi.py
+++ b/api/src/shipit_api/admin/xpi.py
@@ -9,7 +9,7 @@ from taskcluster.exceptions import TaskclusterRestFailure
 from werkzeug.exceptions import BadRequest
 
 from backend_common.taskcluster import get_root_url
-from shipit_api.admin.api import do_schedule_phase, notify_via_matrix
+from shipit_api.admin.api import do_schedule_phase, get_signoff_emails, notify_via_matrix
 from shipit_api.admin.github import get_xpi_type
 from shipit_api.admin.tasks import UnsupportedFlavor, generate_phases
 from shipit_api.common.config import SCOPE_PREFIX
@@ -80,23 +80,17 @@ def get_phase(name, phase):
 
 
 def schedule_phase(name, phase):
-    additional_shipit_emails = set()
-    phase_to_schedule = None
     session = current_app.db.session
     phases = session.query(XPIPhase).filter(XPIRelease.id == XPIPhase.release_id).filter(XPIRelease.name == name).all()
+    phase_to_schedule = list(filter(lambda _phase: _phase.name == phase, phases))
 
     # Get email for all signoffs from previous phases and phase scheduler
-    if phases:
-        for _phase in phases:
-            if _phase.name == phase:
-                phase_to_schedule = _phase
-            if _phase.completed_by is not None:
-                additional_shipit_emails.add(_phase.completed_by)
-            additional_shipit_emails.update({signoff.completed_by for signoff in _phase.signoffs if signoff.completed_by is not None})
+    additional_shipit_emails = get_signoff_emails(phases)
 
     if not phase_to_schedule:
         abort(404, f"phase {phase} not found")
 
+    phase_to_schedule = phase_to_schedule[0]
     # we must require scope which depends on XPI type
     xpi_type = _xpi_type(phase_to_schedule.release.revision, phase_to_schedule.release.xpi_name)
     required_permission = f"{SCOPE_PREFIX}/schedule_phase/xpi/{xpi_type}/{phase_to_schedule.name}"
@@ -104,8 +98,6 @@ def schedule_phase(name, phase):
         user_permissions = ", ".join(current_user.get_permissions())
         abort(401, f"required permission: {required_permission}, user permissions: {user_permissions}")
 
-    # remove duplicates
-    additional_shipit_emails = list(additional_shipit_emails)
     scheduled_phase = do_schedule_phase(session, phase_to_schedule, additional_shipit_emails)
     url = taskcluster_urls.ui(get_root_url(), f"/tasks/groups/{scheduled_phase.task_id}")
     logger.info("Phase %s of %s started by %s. - %s", scheduled_phase.name, scheduled_phase.release.name, scheduled_phase.completed_by, url)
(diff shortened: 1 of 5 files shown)


api/src/shipit_api/admin/xpi.py
```diff
@@ -80,21 +80,37 @@ def get_phase(name, phase):
...
         abort(401, f"required permission: {required_permission}, user permissions: {user_permissions}")
 
-    phase = do_schedule_phase(session, phase)
-    url = taskcluster_urls.ui(get_root_url(), f"/tasks/groups/{phase.task_id}")
-    logger.info("Phase %s of %s started by %s. - %s", phase.name, phase.release.name, phase.completed_by, url)
-    notify_via_matrix("xpi", f"Phase {phase.name} of {phase.release.name} started by {phase.completed_by}. - {url}")
-    return phase.json
+    # remove duplicates
+    additional_shipit_emails = list(set(additional_shipit_emails))
+    scheduled_phase = do_schedule_phase(session, phase_to_schedule, additional_shipit_emails)
```
sarah-clements, line 109: I found it confusing that phase was re-declared in a few places (and even overwrite the phase arg), so I took the liberty of creating additional variables :)

api/src/shipit_api/admin/xpi.py
```diff
@@ -80,21 +80,37 @@ def get_phase(name, phase):
...
     if not current_user.has_permissions(required_permission):
         user_permissions = ", ".join(current_user.get_permissions())
         abort(401, f"required permission: {required_permission}, user permissions: {user_permissions}")
 
-    phase = do_schedule_phase(session, phase)
-    url = taskcluster_urls.ui(get_root_url(), f"/tasks/groups/{phase.task_id}")
-    logger.info("Phase %s of %s started by %s. - %s", phase.name, phase.release.name, phase.completed_by, url)
-    notify_via_matrix("xpi", f"Phase {phase.name} of {phase.release.name} started by {phase.completed_by}. - {url}")
-    return phase.json
+    # remove duplicates
```
sarah-clements, line 107: It's occurred to me that I'm not sure if there is a check to remove duplicates in the staging-xpi-manifests pr, so I'll make a note to look into it.

api/src/shipit_api/admin/xpi.py
```diff
@@ -80,21 +80,37 @@ def get_phase(name, phase):
...
 
 def schedule_phase(name, phase):
+    additional_shipit_emails = []
+    phase_to_schedule = None
     session = current_app.db.session
-    phase = session.query(XPIPhase).filter(XPIRelease.id == XPIPhase.release_id).filter(XPIRelease.name == name).filter(XPIPhase.name == phase).first_or_404()
+    phases = session.query(XPIPhase).filter(XPIRelease.id == XPIPhase.release_id).filter(XPIRelease.name == name).all()
+
+    # Get email for all signoffs from previous phases and phase scheduler
+    if len(phases):
```
bhearsum, line 89: nit: `if phases` would be preferable here

api/src/shipit_api/admin/xpi.py
```diff
@@ -80,21 +80,37 @@ def get_phase(name, phase):
 
 
 def schedule_phase(name, phase):
+    additional_shipit_emails = []
```
bhearsum, line 83: I see you're converting this to a set later - you should just do that up here...

api/src/shipit_api/admin/xpi.py
```diff
@@ -80,21 +80,37 @@ def get_phase(name, phase):
...
+    phases = session.query(XPIPhase).filter(XPIRelease.id == XPIPhase.release_id).filter(XPIRelease.name == name).all()
+
+    # Get email for all signoffs from previous phases and phase scheduler
+    if len(phases):
+        for _phase in phases:
+            if _phase.name == phase:
+                phase_to_schedule = _phase
+            if _phase.completed_by is not None:
+                additional_shipit_emails.append(_phase.completed_by)
+            additional_shipit_emails = additional_shipit_emails + [signoff.completed_by for signoff in _phase.signoffs if signoff.completed_by is not None]
```
bhearsum, line 95: ...and either use `set.add()` in a loop, or `set.update` to do this.

api/src/shipit_api/admin/xpi.py
bhearsum, line 109: ...you could also update `do_schedule_phase` to support a set instead of requiring a list, or do the list conversion here. (It may already Just Work with a set, it's hard to tell by reading the code.)

bhearsum: This is a good readability improvement, thanks!

bhearsum (changes requested):
> Ben, it'd be good to have some tests for this. Where would you prefer I add them?

We've got tests for Ship It in https://github.com/mozilla-releng/shipit/tree/main/api/tests already - you can add them there.

api/src/shipit_api/admin/xpi.py
sarah-clements: Why is it better? :)

api/src/shipit_api/admin/xpi.py
sarah-clements: Sure, I can start with a set but I still need to cast it to a list - empty or not - later on.

api/src/shipit_api/admin/xpi.py
bhearsum: Oh sorry - it's just a bit more pythonic IMO.

api/src/shipit_api/admin/xpi.py
bhearsum: Yeah, for sure. Starting with a set means you're never in a state where you can have duplicate entries though, which may help prevent bugs later.

api/src/shipit_api/admin/xpi.py
sarah-clements: I think I'd prefer to keep the list conversion here.

api/src/shipit_api/admin/xpi.py
sarah-clements: ok

api/src/shipit_api/admin/xpi.py
sarah-clements: Makes sense :)

sarah-clements:
@bhearsum I made changes based on feedback and added a unit test (held off on an integration test for now). Can you take another look please?

bhearsum (approved):
Looks good, thank you!

PR #825, opened 2021-07-28
fix .rstrip() error in xpi name parsing

jmaher:


api/src/shipit_api/admin/github.py
```diff
@@ -261,7 +261,9 @@ def extract_github_repo_owner_and_name(url):
     repo_owner = parts[0]
     repo_name = parts[1]
 
-    return repo_owner, repo_name.rstrip(".git")
+    if repo_name.endswith('.git'):
+      repo_name = repo_name.rstrip('.git')
```
escapewindow, line 265: I think we'll still have a bug: if we have a `spot.git` we'll end up with `spo`.

```
#!/usr/bin/env python
import re
repo = "spot.git"
print(repo.rstrip(".git"))  # spo
print(re.sub("\.git$", "", repo))  # spot
```

escapewindow (changes requested):

api/src/shipit_api/admin/github.py
escapewindow: also linting :) looks like we want 4 spaces

jmaher: I am not aware of a straightforward .trim() or function to pull a string out.  Using the logic of .endwsith(.git), we could do:
repo = repo[:-4]

that removes the last 4 characters

escapewindow: sure, that works. the `re.sub` also works, though regular expressions are sometimes controversial.

escapewindow (approved):
Thanks!

PR #829, opened 2021-07-30
issue 828 - xpi commit link points to wrong repo/revision

jmaher:
not a real fix, but this works and shows where the problem is

### DIFF
--- a/frontend/src/components/ReleaseProgress/index.jsx
+++ b/frontend/src/components/ReleaseProgress/index.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useContext } from 'react';
+import React, { useState, useContext, useEffect } from 'react';
 import Box from '@material-ui/core/Box';
 import Button from '@material-ui/core/Button';
 import Typography from '@material-ui/core/Typography';
@@ -19,6 +19,7 @@ import ReleaseContext from '../../utils/ReleaseContext';
 import { AuthContext } from '../../utils/AuthContext';
 import config from '../../config';
 import { repoUrlBuilder } from '../../utils/helpers';
+import { getXpis } from '../vcs';
 
 const useStyles = makeStyles(() => ({
   cardActions: {
@@ -34,6 +35,7 @@ export default function ReleaseProgress({
   const classes = useStyles();
   const authContext = useContext(AuthContext);
   const mutable = authContext.user && !readOnly;
+  const [xpis, fetchXpis] = useAction(getXpis);
   const [open, setOpen] = useState(false);
   const [cancelState, cancelAction] = useAction(cancelReleaseAPI);
   const { fetchReleases } = useContext(ReleaseContext);
@@ -42,6 +44,18 @@ export default function ReleaseProgress({
   const handleClose = () => {
     setOpen(false);
   };
+
+  const initXpis = async () => {
+    const { owner, project } = config.XPI_MANIFEST;
+
+    await fetchXpis(owner, project, 'HEAD');
+  };
+
+  useEffect(() => {
+    if (config.XPI_MANIFEST) {
+      initXpis();
+    }
+  }, []);
 
   const cancelRelease = async releaseName => {
     const result = await cancelAction(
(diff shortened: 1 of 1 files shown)


frontend/src/components/ReleaseProgress/index.jsx
```diff
@@ -113,7 +113,11 @@ export default function ReleaseProgress({
     }
 
     if (isXPI && XPI_MANIFEST.repo) {
-      url = repoUrlBuilder(XPI_MANIFEST.repo, release.revision);
+      trimmedRevision = release.xpi_revision.substring(0, 13);
+      url = repoUrlBuilder(
+        `https://github.com/mozilla-extensions/${release.xpi_name}`,
```
wagnerand, line 118: For the record, we currently have at least one repository outside of `mozilla-extensions`.

jmaher:
yes, this doesn't work for my normal test of https://github.com/mozilla-releng/staging-xpi-public either.  This is just highlighting where the problem is.  I am not sure of the fix, but can bang on it next week, or maybe someone else has a quick win here.

frontend/src/components/ReleaseProgress/index.jsx
jmaher: I have updated the code to load the repo from the config file

frontend/src/components/ReleaseProgress/index.jsx
```diff
@@ -112,8 +126,38 @@ export default function ReleaseProgress({
...
+        let owner = '';
+
+        xpis.data.xpis.forEach(xpi => {
+          if (xpi.xpi_name === release.xpi_name) {
+            repo = xpi.repo;
+            owner = xpi.owner;
+          }
+        });
+
+        // xpis.data.xpis.forEach(xpi => (repo = xpi.repo));
```
sarah-clements, line 154: Remove this

frontend/src/components/ReleaseProgress/index.jsx
```diff
@@ -93,8 +107,8 @@ export default function ReleaseProgress({
   const renderReleaseTitle = (isXPI, release) => {
     let url = null;
     let productBranch = null;
-    const { PRODUCTS, XPI_MANIFEST, TREEHERDER_URL } = config;
-    const trimmedRevision = release.revision.substring(0, 13);
+    const { PRODUCTS, TREEHERDER_URL } = config;
```
sarah-clements, line 110: Perhaps cleaner to just import the variables we need on line 20 rather than the entire config file - `import { PRODUCTS, XPI_MANIFEST, TREEHERDER_URL } from '../../config'`

frontend/src/components/ReleaseProgress/index.jsx
```diff
@@ -112,8 +126,38 @@ export default function ReleaseProgress({
...
+    if (isXPI) {
+      if (xpis.loading) {
+        return <Spinner loading />;
+      }
+
+      if (xpis.error) {
+        return (
+          <Typography variant="h6" component="h3">
+            {xpis.error.toString()}
+          </Typography>
```
sarah-clements, line 138: I'm wondering if it'd be more useful to not return anything if the request fails and rather report it to sentry.

sarah-clements (approved):
I think this looks good. I appreciate you fixing my mistake!

frontend/src/components/ReleaseProgress/index.jsx
jmaher: unfortunately this is referencing data that is imported into config, I am sure there is a better explanation, but what should be straightforward import X from config is not working.

frontend/src/components/ReleaseProgress/index.jsx
jmaher: done

frontend/src/components/ReleaseProgress/index.jsx
jmaher: I am not clear on how to use SENTRY, there is references to SENTRY_DSN, but I don't see things using it; I did copy this from the newxpi page, so we could make a cleanup in the future.

PR #859, opened 2021-09-09
remove references to deprecated addons type 'normandy'

jmaher:


jmaher:
While this cleans up things, I do wonder if there will be issues to existing addons that might be displayed in shipit UI whose type is now removed.  I don't know if anything in the shipit DB needs to be updated.

escapewindow:
> While this cleans up things, I do wonder if there will be issues to existing addons that might be displayed in shipit UI whose type is now removed. I don't know if anything in the shipit DB needs to be updated.

Hm, is this something we can test in stage before we merge here?

escapewindow (approved):
r=me. I do recommend testing in stage and/or local dev before we merge this PR.

PR #893, opened 2021-09-28
Initial support for pinebuild

gbrownmozilla:
This is very preliminary: I've searched for "DevEdition" and duplicated or expanded code for "PineBuild". 

tox tests pass, and startup looks okay, but I haven't tested further.

### DIFF
--- a/api/src/shipit_api/admin/api.py
+++ b/api/src/shipit_api/admin/api.py
@@ -70,7 +70,7 @@ def add_release(body):
     session = current_app.db.session
     partial_updates = body.get("partial_updates")
     if partial_updates == "auto":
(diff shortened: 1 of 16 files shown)


escapewindow (changes requested):
This looks good! However, when trying to create a maple Pinebuild beta on [shipit stage](https://shipit.staging.mozilla-releng.net/new), I get the following error:

Screen Shot 2021-09-29 at 15 00 21

Do we know why this is?

gbrownmozilla:
> Do we know why this is?

Ben found this error in sentry: Cannot bump "patch_number". New version number is not valid. Cause: "94.0.1a1" does not match the pattern: Patch number cannot be defined on a nightly version

escapewindow (approved):

PR #954, opened 2021-11-22
Add Focus to frontend configs

sarah-clements:
Tested it locally (just seeing that the staging repo came up when trying to create a new release). 

Edit: Not ready yet, just realized I might have forgotten some parts :)

### DIFF
--- a/api/src/shipit_api/admin/release.py
+++ b/api/src/shipit_api/admin/release.py
@@ -6,9 +6,10 @@
 import logging
 
 import requests
-from mozilla_version.fenix import FenixVersion  # TODO Move FenixVersion into mozilla.gecko
+from mozilla_version.fenix import FenixVersion  # TODO replace with MobileVersion
 from mozilla_version.gecko import DeveditionVersion, FennecVersion, FirefoxVersion, ThunderbirdVersion
-from mozilla_version.maven import MavenVersion  # TODO To support android-components
+from mozilla_version.maven import MavenVersion
+from mozilla_version.mobile import MobileVersion
 
 from shipit_api.common.config import SUPPORTED_FLAVORS
 from shipit_api.common.product import Product, get_key
--- a/api/src/shipit_api/admin/release.py
+++ b/api/src/shipit_api/admin/release.py
@@ -23,6 +23,7 @@ _VERSION_CLASS_PER_PRODUCT = {
     Product.FENIX: FenixVersion,
     Product.FENNEC: FennecVersion,
     Product.FIREFOX: FirefoxVersion,
+    Product.FOCUS_ANDROID: MobileVersion,
     Product.THUNDERBIRD: ThunderbirdVersion,
 }
 
--- a/api/src/shipit_api/admin/settings.py
+++ b/api/src/shipit_api/admin/settings.py
@@ -74,6 +74,7 @@ LDAP_GROUPS = {
     "fenix-signoff": ["shipit_mobile"],
     "android-components-signoff": ["shipit_mobile"],
     "thunderbird-signoff": ["shipit_thunderbird"],
+    "focus-android-signoff": ["shipit_mobile"],
     # XPI signoffs. These are in flux.
     # Adding Releng as a backup to most of these, for bus factor. Releng should
     # only sign off if requested by someone in the appropriate group.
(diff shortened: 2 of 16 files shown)


jmaher:
I don't know what else to add here- the changes you are making look good though- when you are ready for review I am happy to look more.

api/src/shipit_api/admin/release.py
```diff
@@ -9,6 +9,7 @@
 from mozilla_version.fenix import FenixVersion  # TODO Move FenixVersion into mozilla.gecko
 from mozilla_version.gecko import DeveditionVersion, FennecVersion, FirefoxVersion, ThunderbirdVersion
 from mozilla_version.maven import MavenVersion  # TODO To support android-components
+from mozilla_version.mobile import MobileVersion
```
jmaher, line 12: we have these separate modules like fenix, maven, mobile.  It is probably fine, but the TODO comments that exist for fenix and maven confuse me- is there more todo here?  should we do the right thing if it is to move into gecko, or are we doing the right thing already and the TODO statements should go away?

api/src/shipit_api/admin/release.py
```diff
@@ -22,6 +23,7 @@
     Product.FENIX: FenixVersion,
     Product.FENNEC: FennecVersion,
     Product.FIREFOX: FirefoxVersion,
+    Product.FOCUS_ANDROID: MobileVersion,
```
jmaher, line 26: I also noticed that we have FENNEC throughout, AIUI, we don't ship fennec and haven't for the last year or so.

frontend/src/configs/production.js
```diff
@@ -164,6 +164,25 @@ module.exports = {
...
+        {
+          branch: '',
+        },
+      ],
+      repositories: [
+        {
+          prettyName: 'Official repo',
+          project: 'focus-android',
+          repo: 'https://github.com/mozilla-mobile/focus-android',
+          enableReleaseEta: false,
```
jmaher, line 181: this is production, do we want this to be True?

api/src/shipit_api/admin/release.py
sarah-clements: The MavenVersion comment is outdated since its already used for android-components - will remove that. I've made a MobileVersion that's a [dupe of FenixVersion](https://github.com/mozilla-releng/mozilla-version/pull/89), so I was thinking I'd replace FenixVersion use with MobileVersion in a separate pr (but I can update the comment in the meantime).

api/src/shipit_api/admin/release.py
sarah-clements: I'm wondering if it might be needed for historical reasons? Can look into it more and file a separate cleanup pr.

frontend/src/configs/production.js
sarah-clements: Good question... looks like we only enable that for Firefox releases. I just copied over what was done for fenix. Looks like its used to allow a scheduling component to be shown in the UI, so this is probably something we can change later if it should in fact be enabled.

sarah-clements:
I think is is ready for final review. You can see the focus for android test releases that I created on staging: https://shipit.staging.mozilla-releng.net/

api/src/shipit_api/admin/settings.py
```diff
@@ -74,6 +74,7 @@
     "fenix-signoff": ["shipit_mobile"],
     "android-components-signoff": ["shipit_mobile"],
     "thunderbird-signoff": ["shipit_thunderbird"],
+    "focus-android-signoff": ["shipit_mobile"],
```
jmaher, line 77: are we sure the same people who sign off on Fenix == sign off on focus android?  This is really relman:
ryanvm
pchevrel
telin
dsmith

api/tox.ini
```diff
@@ -26,7 +26,7 @@ usedevelop = true
 deps =
     -r {toxinidir}/requirements/test.txt
 commands =
-    {posargs:py.test --cov-config=tox.ini --cov-append --cov=shipit_api --cov-report term-missing tests}
+    {posargs:py.test -vv --cov-config=tox.ini --cov-append --cov=shipit_api --cov-report term-missing tests}
```
jmaher, line 29: do we want -vv normally?  not sure if this was debugging stuff locally or an intended change.

jmaher (commented):
2 new comments I would like to see addressed before r+, overall really close

api/tox.ini
sarah-clements: Its an intended change because it shows more in the logs and locally when a test fails. I think its more helpful to show the full diff for failed tests instead of a little snippet. But I don't feel strongly about it and so I can remove it if preferred.

api/src/shipit_api/admin/settings.py
sarah-clements: I got a confirmation from RyanVM focus signoffs should be the same as for Fenix and A-C.

jmaher (approved):
thanks Sarah!  let's get this landed

PR #955, opened 2021-11-23
upgrade python requires and to 3.9

jmaher:


jmaher:
this is basically:
 * pip-compile-multi
 * fixing the changes to connexion/flask stuff for error handling (copied the logic from flask as it was not a function, but moved to a method in FlaskApp which we were not using
 * connexion upgrade is not connexion2 and swaggerUI is now openapi.
 * upgrade to 3.9.7 to be like many other tools- why not!
 
tested locally and on dev:
https://shipit.staging.mozilla-releng.net/xpi

I specifically looked for CORS, Internal server errors, and error handling.  I was able to test these conditions and general use of shipit to add/edit/cancel releases locally and on dev: https://shipit.staging.mozilla-releng.net/.  

Happy to try other things if some of this looks too scary or large of a change for a single push.

sarah-clements (approved):
Looks okay to me.

PR #957, opened 2021-11-26
Change config to ignore non-security dot releases

sarah-clements:
I did the same thing for Balrog and will hopefully make these updates a bit more manageable.


jmaher (approved):
this should reduce the spam significantly, thanks

PR #964, opened 2021-12-10
pin python version to 3.8.12 in docker-image kind

sarah-clements:
Testing

### DIFF
--- a/taskcluster/ci/docker-image/kind.yml
+++ b/taskcluster/ci/docker-image/kind.yml
@@ -14,7 +14,7 @@ jobs:
     python38:
         definition: python
         args:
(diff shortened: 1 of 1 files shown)


jmaher (approved):
thanks.  I have been moving everything else to 3.9.7, not asking you to change here, but it might be worth an upgrade or an issue in the future.

sarah-clements:
We don't actually need this, was more of an experiment but I solved one of the issues with my pr. And yeah, we really should update the ci docker images and tests to run with python39, especially since that's what tox uses locally!

sarah-clements:
Will close this out in favor of upgrading the CI to use 3.9

PR #967, opened 2021-12-14
fix malformed XPIRelease openapi schema

gabrielBusta:
fix malformed XPIRelease openapi schema in api/src/shipit_api/admin/api.yml


jmaher (approved):
thanks for the cleanup

PR #968, opened 2021-12-16
Expose artifacts in shipit

gabrielBusta:
Link to xpi packages from shipit.

### DIFF
--- a/api/src/shipit_api/admin/tasks.py
+++ b/api/src/shipit_api/admin/tasks.py
@@ -55,6 +55,33 @@ def find_decision_task_id(repo_url, project, revision, product):
         raise Exception(f"route {decision_task_route} exception {exc}")
 
 
+@lru_cache(maxsize=2048)
+def find_task_group(task_id):
+    queue = get_service("queue")
+    try:
+        return queue.listTaskGroup(task_id)
+    except Exception as exc:
+        raise Exception(f"task {task_id} exception {exc}")
+
+
+@lru_cache(maxsize=2048)
+def find_latest_artifacts(task_id):
+    queue = get_service("queue")
+    try:
+        return queue.listLatestArtifacts(task_id)['artifacts']
+    except Exception as exc:
+        raise Exception(f"task {task_id} exception {exc}")
+
+
+@lru_cache(maxsize=2048)
+def generate_artifact_url(task_id, artifact):
+    queue = get_service("queue")
+    try:
+        return queue.buildUrl("getLatestArtifact", task_id, artifact)
+    except Exception as exc:
+        raise Exception(f"task {task_id} exception {exc}")
+
+
 def fetch_artifact(task_id, artifact):
     try:
         queue = get_service("queue")
--- a/api/src/shipit_api/admin/tasks.py
+++ b/api/src/shipit_api/admin/tasks.py
@@ -3,9 +3,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import os
 import copy
 import json
+import os
 from functools import lru_cache
 
 import jsone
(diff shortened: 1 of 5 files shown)


api/src/shipit_api/admin/tasks.py
```diff
@@ -55,6 +56,47 @@ def find_decision_task_id(repo_url, project, revision, product):
         raise Exception(f"route {decision_task_route} exception {exc}")
 
 
+def fetch_group_tasks(task_id):
+    queue = get_service("queue")
+    try:
+        return queue.listTaskGroup(task_id)["tasks"]
```
escapewindow, line 62: We may want to add a retry around this, e.g. wrapping with [redo](https://pypi.org/project/redo/), since sometimes the queue will hiccup.

api/src/shipit_api/admin/tasks.py
```diff
@@ -55,6 +56,47 @@ def find_decision_task_id(repo_url, project, revision, product):
...
+
+    Args:
+        task_id: The task id for the action task associated with the release phase.
+        xpi_name: The xpi name set on the xpi release.
+
+    Returns:
+        A taskcluster url pointing to a phase's .xpi artifact or the empty string.
+    """
+    tasks = fetch_group_tasks(task_id)
+    for task in tasks:
```
escapewindow, line 89: Does this mean `tasks` is a whole group worth of tasks, and we find the first task and return an artifact url?
If so, we probably need to narrow down by the release-signing kind or label.

If it helps, we could generate indexes for the release-signing tasks, like https://firefox-ci-tc.services.mozilla.com/tasks/index/xpi.v2.normandy-devtools.revision.021f115cb7cd5b7e06193d9ff5c28ff29e791770.taskgraph/decision except pointing at the release-signing task, which may help avoid querying the task group and then going through all the tasks to find the release-signing task.

escapewindow:
It looks like we need to format our code with `black` as well.
I'm curious what this looks like? Should we roll out to shipit-stage?

Overall, great work! Thank you! Let me know if you have questions about my comments.

api/src/shipit_api/admin/tasks.py
gabrielBusta: Yea it's the group of tasks spawned by the decision/action task id corresponding to the release phase.

gabrielBusta:
@escapewindow its a download icon below the progress stage... altho I think it could be a big friendly "Add to Firefox" button at the corner of the release. Inspired by the one in about:addons tab in Firefox and [addons.mozilla.org](https://addons.mozilla.org/en-US/firefox/addon/ublock-origin/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=search)

I will look into adding the index to pull up the relevant task directly.

frontend/src/components/PhaseProgress/index.jsx
```diff
@@ -145,34 +147,15 @@ export default function PhaseProgress({ release, readOnly, xpi }) {
...
-                  <React.Fragment>
-                    <Typography variant="caption" display="block">
-                      Add <b>{release.name}</b> to Firefox
-                    </Typography>
-                  </React.Fragment>
-                }
-              />
+                position="absolute">
+                <Link href={phase.xpiUrl}>
+                  {release.xpi_name}.xpi
```
jmaher, line 156: is xpi_name and xpiUrl available all the time here?  We have firefox releases and xpi releases in different pages.  Test this on all pages

gabrielBusta: because of [this](https://github.com/mozilla-releng/shipit/pull/968/files#diff-4e76cbc004b97aebb98430c5b53465ffb9a0d59235898dfa9fa186487427d5cfR149) condition it will render this link only when there's an xpi url in the phase payload. When there's an xpi url, there is always an xpi name

jmaher (approved):
looks good

jmaher (commented):
looks good to me

jmaher (commented):
this was too easy

PR #973, opened 2021-12-21
xpi signoff flag

gabrielBusta:
xpi signoff flag to make local testing easier, without having to fiddle with code.


jmaher (approved):
thanks!

PR #977, opened 2021-12-22
fix xpi package link

gabrielBusta:
fix xpi package link


jmaher (approved):
oh fantastic- this is more robust!

PR #986, opened 2022-01-11
Update xpi-manifest branch to main

hneiva:


jmaher (approved):
thanks- I really like reviewing small patches.  For reference, how do people who have xpi-manifest checked out update their local development/upstream branches?  Can we also name the master branch read only?

escapewindow:
> thanks- I really like reviewing small patches. For reference, how do people who have xpi-manifest checked out update their local development/upstream branches? Can we also name the master branch read only?

I have the opposite thought -- keeping the old branch around encourages people to use it, and they base their changes on super crufty revisions.

PR #990, opened 2022-01-20
frontend: don't filter releases based on branch when guessing build number

jcristau:
The release name, i.e. (product, version, build_number) tuple, must be
unique, so if we filter on branch we can fail at submit time with a
constraint violation.

This shouldn't matter on prod because the version/branch mapping is
unambiguous but on staging this can happen (different github repo
branches, or hg project repos).

### DIFF
--- a/frontend/src/components/api.js
+++ b/frontend/src/components/api.js
@@ -10,9 +10,9 @@ import config from '../config';
  * This will fetch all release builds including shipped, aborted, and not
  * started yet.
  */
-export async function getBuildNumbers(product, branch, version) {
+export async function getBuildNumbers(product, version) {
   const res = await axios.get('/releases', {
-    params: { product, branch, version, status: 'shipped,aborted,scheduled' },
+    params: { product, version, status: 'shipped,aborted,scheduled' },
     usePublicApi: true,
   });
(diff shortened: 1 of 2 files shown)


frontend/src/components/api.js
```diff
@@ -12,7 +12,7 @@ import config from '../config';
  */
 export async function getBuildNumbers(product, branch, version) {
```
jcristau, line 13: We could drop the branch arg here entirely, if that's preferred.

jmaher: what are the risks of removing branch?  will we get a larger list of build numbers to choose from?  Can we do this conditionally if the branch is unknown?

if we do decide to get rid of the branch from the params, we should remove it from the function definition.

jcristau: The risk would be that in theory we can then schedule build1 from one branch and build2 from the other.  I don't see how we could do it conditionally.

jmaher: ok, then please remove `branch` from the parameters

gabrielBusta (approved):

gbrownmozilla (approved):

PR #1006, opened 2022-02-08
reenable normandy-privileged

escapewindow:
Largely reverts #859, but leaves out the nonexistent?
`xpi_normandy_admin` ldap group.

We have two problems: first, we had abandoned normandy-privileged
releases in-flight in production shipit, and after landing #859 we no
longer had scopes to either move forward or cancel those releases
without either reverting or reaching into the DB.

Second, we have an unexpected normandy-privileged release coming, and
there's [pushback](https://github.com/mozilla-extensions/xpi-manifest/pull/149#discussion_r801858020) against using the `privileged` format for this,
even if they both use the exact same signing format behind the scenes.

Let's reenable `normandy-privileged` and leave it enabled until it's
clear we will definitely not ship any more normandy-privileged addons.

### DIFF
--- a/api/src/shipit_api/common/config.py
+++ b/api/src/shipit_api/common/config.py
@@ -530,5 +530,13 @@ SIGNOFFS = {
                 },
             ],
         },
+        "normandy-privileged": {
+            "promote": [
+                {"name": "Normandy privileged admin", "description": "Promote XPI", "permissions": "xpi_normandy-privileged_signoff"},
+            ],
+            "ship": [
+                {"name": "Normandy privileged admin", "description": "Ship XPI", "permissions": "xpi_normandy-privileged_signoff"},
+            ],
+        },
     },
 }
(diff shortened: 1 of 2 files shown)


api/src/shipit_api/common/config.py
```diff
@@ -530,5 +530,13 @@
                 },
             ],
         },
+        "normandy-privileged": {
+            "promote": [
+                {"name": "Normandy privileged admin", "description": "Promote XPI", "permissions": "xpi_normandy-privileged_signoff"},
+            ],
+            "ship": [
+                {"name": "Normandy privileged admin", "description": "Ship XPI", "permissions": "xpi_normandy-privileged_signoff"},
```
jmaher, line 538: do we need `release` ?  not for cleanup, but maybe the new release?  I know they will not release to but will want our signing.

jmaher (approved):
we will also need to update xpi-template to allow normandy[-privileged] if we are going to build.  I guess we can verify this cleans things up

api/src/shipit_api/common/config.py
escapewindow: Trying to figure out what this means :)
I think you mean `promote` to sign, but I may be missing something. Maybe I'll figure out during testing :)

jmaher: yeah, I think I was a bit incorrect in my statement.  we need `promote` to sign, which is part of this patch.

PR #1011, opened 2022-02-16
product-details: rollback the db session when we're done with it

jcristau:
The worker and its db session are long lived, so if one rebuild hits an
error when querying the db, it appears we leave the session in
pending-rollback state and never recover.


escapewindow (approved):
Seems to match https://stackoverflow.com/a/33284980 , so r+ :)

PR #1052, opened 2022-04-27
Update some python deps

gbrownmozilla:
I will try to update additional deps in future PRs. These seemed pretty minor - a safe batch to start with.


gabrielBusta (approved):
This is a good start :)

Issue #1066, opened 2022-05-11
Set ESR_NEXT to 102


jcristau:
We need to do the same as #807 for this new ESR cycle.

cc @gabrielBusta @gbrownmozilla

PR #807, opened 2021-07-12
Bug 1717527 - add ESR91

jcristau:
As far as I can tell it should be safe to set ESR_NEXT already.  There are 2 users, in `api/src/shipit_api/admin/product_details.py`:
- `get_product_categories` won't find anything that matches `91(\.[0-9]+){1,2}esr$` until the 91.0esr release
- `get_firefox_esr_next_version`'s call into `get_latest_version` won't find anything with `branch` set to `releases/mozilla-esr91`, and thus it will still return an empty string, until 91.0esr ships

PR #1069, opened 2022-05-11
Bug 1766696 - Add ESR102

gabrielBusta:
https://github.com/mozilla-releng/shipit/issues/1066


gbrownmozilla (approved):
This appears to accurately parallel https://github.com/mozilla-releng/shipit/pull/807/files.

jcristau (approved):
Thanks!

We might want to remove esr78 from the frontend config too, that seems overdue.

PR #1080, opened 2022-05-30
add pin.sh

gbrownmozilla:
Like https://github.com/mozilla-releng/balrog/pull/2533, for shipit now.

### DIFF
--- a/api/maintenance/pin-helper.sh
+++ b/api/maintenance/pin-helper.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+# This runs in docker to pin our requirements files.
+set -e
+SUFFIX=${SUFFIX:-txt}
+if [ $# -gt 0 ]; then
+    EXTRA_PCM_ARGS="$@"
+fi
+
+pip install --upgrade pip
+pip install pip-compile-multi
+
+apt-get update
+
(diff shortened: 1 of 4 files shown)


gbrownmozilla:
@gabrielBusta review ping?

api/maintenance/pin-helper.sh
```diff
@@ -0,0 +1,16 @@
...
+set -e
+SUFFIX=${SUFFIX:-txt}
+if [ $# -gt 0 ]; then
+    EXTRA_PCM_ARGS="$@"
+fi
+
+pip install --upgrade pip
+pip install pip-compile-multi
+
+apt-get update
```
hneiva, line 12: As a mac user, I don't love that we hard-code this apt-command here. There are pin scripts on other projects that run natively without the need of docker.
Unless if this is truly necessary, I'd suggest removing it. As I see it, the docker image "should" already be updated enough for this job.

hneiva (approved):

api/maintenance/pin-helper.sh
gbrownmozilla: The apt-get update is there only by way of copy/paste from other projects' pin scripts. I agree it does seem extraneous...let's see if removing it causes any trouble in the long term.

PR #1087, opened 2022-05-31
Mark phase as completed when the task graph has completed

gabrielBusta:
Modifies the ShipIt front-end to display the Phase state taking into account the state of the tasks in the graph's group.

### DIFF
--- a/frontend/src/components/PhaseProgress/index.jsx
+++ b/frontend/src/components/PhaseProgress/index.jsx
@@ -177,7 +177,8 @@ export default function PhaseProgress({ release, readOnly, xpi }) {
       (idx === 0 || // The first phase can be scheduled anytime
       allowPhaseSkipping || // Can schedule anything
       phases[idx - 1].tcStatus === 'completed' || // previsous phase is done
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -80,7 +80,7 @@ services:
 
   ui:
     # SSL doesn't work with node:current
-    image: node:10
+    image: node:16
```
gabrielBusta, line 83: Why is this showing in the diff? The code in main has 16 https://github.com/mozilla-releng/shipit/blob/main/docker-compose.yml#L83

gabrielBusta: fixed it

gbrownmozilla (approved):
Looks good to me!

gabrielBusta:
Modified it a bit so it won't block Gecko releases if there are exceptions / failures in the task graph

gabrielBusta:
[If we do this we only block on add on release task group failures](https://github.com/mozilla-releng/shipit/commit/032adf1f60e61c8a02a7ae151640325aa2ff8080)

PR #1088, opened 2022-06-01
Update local node version to 16

gabrielbusta:
Fixes
```
shipit-ui-1 | error is-absolute-url@4.0.1: The engine "node" is incompatible with this module. Expected version "^12.20.0 || ^14.13.1 || >=16.0.0". Got "10.24.1
```


gbrownmozilla (approved):
Thanks!

PR #1111, opened 2022-07-04
Don't use mutable objects as default arguments

jcristau:
It doesn't look like there's an actual bug here but it's bad form and a
recipe for subtle bugs.

### DIFF
--- a/api/src/cli_common/command.py
+++ b/api/src/cli_common/command.py
@@ -27,7 +27,7 @@ def hide_secrets(text, secrets):
     return text
 
 
-def run(command, stream=False, handle_stream_line=None, log_command=True, log_output=True, secrets=[], **kwargs):
+def run(command, stream=False, handle_stream_line=None, log_command=True, log_output=True, secrets=(), **kwargs):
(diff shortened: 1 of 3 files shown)


JohanLorenzo (approved):
100% agree on this! For people wondering: parameters in function signatures are global objects. Thus, they're shared between calls. For example:

```py
def foo(bar=[]):
    print(bar)
    bar.append("baz")

foo() # prints []
foo() # prints ['baz']
foo() # prints ['baz', 'baz']

```

That's why we must not set mutable objects as defaut values in function signatures.

gabrielBusta (approved):

gbrownmozilla (approved):
@jcristau How do you find these? Just sharp eyes, or are there tools involved?

jcristau:
@gbrownmozilla I noticed one case while looking at the fenix bustage yesterday, then `grep` for similar patterns, pretty much.

I should note I've left one case in `api/src/backend_common/__init__.py` untouched because I couldn't quickly figure out what `app.__extensions` is supposed to be / where it's used.

PR #1112, opened 2022-07-04
Bug 1776830 - part 2 to 4: Code coverage and allow hardcoded flavors to be absent from actions.json

JohanLorenzo:


api/src/shipit_api/admin/tasks.py
```diff
@@ -141,10 +144,13 @@ def extract_our_flavors(avail_flavors, product, version, partial_updates, produc
...
-        description = f"Some flavors are not in actions.json: {all_flavors.difference(set(avail_flavors))}."
-        raise UnsupportedFlavor(description=description)
-    return SUPPORTED_FLAVORS[product_key]
+    avail_flavors_set = set(avail_flavors)
+    if not avail_flavors_set.issuperset(all_flavors):
+        logger.warning(f"Some hardcoded flavors are not in actions.json: {all_flavors.difference(avail_flavors_set)}.")
+    if not avail_flavors_set.issubset(all_flavors):
+        logger.warning(f"Some flavors in actions.json are not hardcoded in shipit: {avail_flavors_set.difference(all_flavors)}.")
+
+    return [flavor for flavor in SUPPORTED_FLAVORS[product_key] if flavor.get("name") in avail_flavors_set]
```
jcristau, line 153: Any reason not to use `flavor["name"]` here?  We use it in `all_flavors` just above.

jcristau (commented):
Thanks!

I *think* this is fine; still trying to think through corner cases...

api/src/shipit_api/admin/tasks.py
JohanLorenzo: Good call,  done!

JohanLorenzo:
Thanks! I'm much more confident with the new implementation now that we have unit tests. Before making the changes in the implementation, shipit:

 * failed if a `shipping_phase` (aka `flavor`) was missing in `actions.json`
 * silently ignored the case where more `shipping_phases` were present in `actions.json`. 

Now, we warn on both cases. On the bright side, we don't return any data coming from `actions.json`. So, no data injection can come from arbitrary tasks. 

Maybe a good addition would be to report the warning on the UI, but that's a bigger patch to implement.

jcristau (approved):

PR #1115, opened 2022-07-06
Bug 1778290 - Disable phase schedule/signoff button once clicked

gabrielBusta:
Mitigates double-scheduling of phases by double-clicking on the "schedule" in the "Schedule Phase" modal.

### DIFF
--- a/frontend/src/components/PhaseProgress/index.jsx
+++ b/frontend/src/components/PhaseProgress/index.jsx
@@ -37,6 +37,9 @@ const useStyles = makeStyles({
 export default function PhaseProgress({ release, readOnly, xpi }) {
   const classes = useStyles();
   const { fetchReleases } = useContext(ReleaseContext);
+  const [disableScheduleOrSignoff, setDisableScheduleOrSignoff] = useState(
+    false
+  );
   const [open, setOpen] = useState(false);
   const [phase, setPhase] = useState({});
   const [selectedSignoffUID, setSelectedSignoffUID] = useState(null);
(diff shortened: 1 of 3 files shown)


gabrielBusta:
Yeah there could be. In those cases we can refresh the page an try again. This to prevent double clicks but we can still double schedule a phase by refreshing and clicking schedule again before it is marked a submitted.

jcristau (approved):
Thanks!

frontend/src/components/PhaseProgress/index.jsx
```diff
@@ -61,6 +64,8 @@ export default function PhaseProgress({ release, readOnly, xpi }) {
   };
 
   const scheduleOrSignoff = async () => {
+    setDisableScheduleOrSignoff(true);
```
JohanLorenzo, line 67: n00b question (I've never written any react code): Here we're disabling the button just before sending the request to the backend. Does react magically re-enable it later? Or the button remains disable until we recreate the modal window?

JohanLorenzo (approved):
Thank you very much for making the fix so fast! I'm no react expert so my review isn't worth a lot here. That said, the patch makes sense to me!

frontend/src/components/PhaseProgress/index.jsx
gabrielBusta: You can re-enable it by refreshing the page. That restarts the app state. I can re-enable it after the request comes back from the back-end.

PR #1119, opened 2022-07-07
Bug 1778571 - Add codecov to ShipIt

gabrielBusta:
These changes enable code coverage reporting via pytest-cov and codecov.io


gabrielBusta:
This is currently set up to fail a check if the PR adds any new uncovered lines. I think that is a good thing, it has prompted me to write tests for taskgraph when working on non-urgent changes. If we need to add an urgent change we can merge PRs even if the task fails, so it is not too annoying

gabrielBusta:
I will land it for now. We can tweak that later depending on how useful it is.

PR #1124, opened 2022-07-11
Fix local Python services

gabrielBusta:
Fix local Python services. The config object can be a `PosixPath` object in py39


gbrownmozilla (changes requested):
This was landed elsewhere. Abandon this one?

PR #1153, opened 2022-09-05
Bug 1789241 - Bump FENIX_NIGHTLY and bring it closer to FIREFOX_NIGHT…

JohanLorenzo:
…LY so they get updated at the same time


jcristau (commented):
Is there ever a case where we'd want FIREFOX_NIGHTLY and FENIX_NIGHTLY to be different?  If not maybe let's just set `FENIX_NIGHTLY = FIREFOX_NIGHTLY` to have one less thing to touch?

JohanLorenzo:
True! During the migration we use to have different version numbers but that shouldn't be the case anymore. Let's use `FIREFOX_NIGHTLY`, then. Thanks for the idea! 👍

jcristau (approved):
Maybe add a comment next to the FIREFOX_NIGHTLY setting to say it's for both desktop and fenix?

gbrownmozilla (approved):
Yeah!

PR #1159, opened 2022-09-12
Expose STRINGFREEZE_DATE in product-details

gabrielBusta:
https://bugzilla.mozilla.org/show_bug.cgi?id=1781016


api/src/shipit_api/common/config.py
```diff
@@ -53,10 +54,13 @@
 # The next 6 dates are information about the current and next release
```
gbrownmozilla, line 54: Can we update the comment to acknowledge the extra lines? Maybe "the 6 constant dates below..."? Or maybe it would be better if DATE_FORMAT, LAST_STRINGFREEZE_DATE, and NEXT_STRINGFREEZE_DATE were all moved below NEXT_RELEASE_DATE??

I'm just trying to keep the regular updates dead-simple.

gbrownmozilla (approved):
Thanks!

PR #1188, opened 2022-10-27
Bug 1786780 - Add firefox-android monorepo

JohanLorenzo:
This needs to be deployed to production on Monday, October 31st.

### DIFF
--- a/frontend/src/configs/dev.js
+++ b/frontend/src/configs/dev.js
@@ -108,9 +108,16 @@ module.exports = {
       ],
       repositories: [
         {
-          prettyName: 'Staging fork',
+          // TODO bug 1797700, remove the staging-android-components repo
+          prettyName: 'Deprecated fork',
           project: 'staging-android-components',
           repo: 'https://github.com/mozilla-releng/staging-android-components',
+          enableReleaseEta: false,
+        },
+        {
+          prettyName: 'Staging Android monorepo',
+          project: 'staging-firefox-android',
+          repo: 'https://github.com/mozilla-releng/staging-firefox-android',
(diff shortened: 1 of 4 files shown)


jcristau (approved):

frontend/src/configs/dev.js
```diff
@@ -108,11 +108,18 @@ module.exports = {
...
-          prettyName: 'Staging fork',
+          // TODO bug 1797700, remove the staging-android-components repo
+          prettyName: 'Deprecated fork',
           project: 'staging-android-components',
           repo: 'https://github.com/mozilla-releng/staging-android-components',
           enableReleaseEta: false,
         },
+        {
+          prettyName: 'Staging Android monorepo',
+          project: 'firefox-android',
```
gbrownmozilla, line 119: Why isn't the project 'staging-firefox-android' (like staging-android-component, staging-focus-android, etc.)? Does it matter?

JohanLorenzo: Good catch! I actually don't know the impact of it because that config worked on the staging instance 🤔 I'll investigate this later. In the meantime, I'm putting `staging-firefox-android` just for the principle of least surprise.

gbrownmozilla (approved):
Thanks.

frontend/src/configs/dev.js
jcristau: From a quick look, it seems this is used to build treeherder URLs, so staging-firefox-android is correct there (but treeherder links for mobile projects don't actually work, because releases are built off of branches but treeherder only looks at commits on trunk)

gabrielBusta (approved):

PR #1192, opened 2022-11-04
Don't show an error on a phase if it is not blocking the following phase

gabrielBusta:
The red triangle of doom 🔺 implies that a phase failed and its blocking the scheduling of the next phase. For Firefox, phases with failed tasks are non-blocking.

### DIFF
--- a/frontend/src/components/api.js
+++ b/frontend/src/components/api.js
@@ -349,6 +349,18 @@ async function getPhaseSignOffs(releaseName, phaseName, url = '/signoff') {
   return req.data.signoffs;
 }
 
+async function getStatus(release, phase) {
+  const isXPIRelease = 'xpi_name' in release;
+
+  if (isXPIRelease) {
+    return getTcStatus(phase);
+  }
+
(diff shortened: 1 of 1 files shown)


gbrownmozilla (approved):

frontend/src/components/api.js
```diff
@@ -366,7 +378,7 @@ export async function getPendingReleases(
           );
 
           if (phase.submitted && phase.actionTaskId) {
-            const tcStatus = await getTcStatus(phase);
+            const tcStatus = await getStatus(release, phase);
```
jcristau, line 381: I think https://github.com/mozilla-releng/shipit/blob/3813c52561aad737ed94f0b8fc02d382e150f602/frontend/src/components/PhaseProgress/index.jsx#L189-L191 should be removed as well, because now the filtering for xpi is done in getStatus

jcristau (approved):

frontend/src/components/api.js
gabrielBusta: I’ll follow up on this later. Landing this to get the fix on main

PR #1203, opened 2022-11-14
Bump config for nightly 109

jcristau:
cc @jfx2006 

[to be deployed on Nov 15]

### DIFF
--- a/api/src/shipit_api/common/config.py
+++ b/api/src/shipit_api/common/config.py
@@ -49,17 +49,17 @@ ESR_BRANCH_PREFIX = "releases/mozilla-esr"
 # e.g. version bumped, but still no builds available.
 #
 # This version also defines the mobile nightly version (i.e.: Fenix)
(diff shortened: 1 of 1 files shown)


gbrownmozilla (approved):

api/src/shipit_api/common/config.py
```diff
@@ -75,7 +75,7 @@
 
 # Thunderbird configs
 LATEST_THUNDERBIRD_ALPHA_VERSION = "54.0a2"
-LATEST_THUNDERBIRD_NIGHTLY_VERSION = "108.0a1"
+LATEST_THUNDERBIRD_NIGHTLY_VERSION = "109.0a1"
```
gbrownmozilla, line 78: There's also #1204

jfx2006: oh oops! I didn't notice that Jcristau took care of Thunderbird already. I'll cancel #1204.

gabrielBusta (approved):

PR #1247, opened 2023-01-05
Bug 1802483 - try not to break the UI when pending releases have expired tasks

jcristau:
This happens easily on staging because try tasks expire quickly, and releases can stay unshipped for extended periods.

### DIFF
--- a/frontend/src/components/api.js
+++ b/frontend/src/components/api.js
@@ -358,6 +358,10 @@ async function getStatus(release, phase) {
 
   const status = await getTaskStatus(phase.actionTaskId);
 
+  if (!status) {
+    return null; // expired?
+  }
+
   return status.status.state;
 }
 


jcristau:
Tested locally:
```
shipitdb=# insert into shipit_api_releases (name, product, version, branch, revision, build_number, status) values ('Firefox-108.0b9-build1', 'firefox', '108.0b9', 'try', '880870afd29ed4a90c838d5b1c7a8a87ad23670a', 1, 'scheduled');
INSERT 0 1
shipitdb=# insert into shipit_api_phases (name, submitted, task_id, task, release_id, context) values ('promote_firefox', true, 'ZmKo1ZlMQYaB6A_uwkFSaA', '', 1, '');
INSERT 0 1
```

Without this change the pending releases page is empty, with it I can see the added release, so that's an improvement.  Probably a bit more work needed to be able to cancel such a release, so I'll look into that as well.

JohanLorenzo (approved):
Thanks for the manual tests! This looks good to me, then!

PR #1259, opened 2023-01-26
frontend: don't truncate the release timestamp creation string

jcristau:


gbrownmozilla (approved):

gabrielBusta (approved):
Nice, I did not like this about the UI

PR #1288, opened 2023-02-24
Fix blocked XPI phases

gabrielBusta:


ahal (approved):

gbrownmozilla (approved):

jcristau:
@gabrielBusta can you maybe explain what this was for and why we no longer want to block on failed tasks?  Presumably this was added on purpose...

PR #1294, opened 2023-03-03
Bug 1820194 - Support `firefox-android` when parsing existing product…

JohanLorenzo:
…-details files

Patch tested locally. I was able to get rebuild product details inside a docker container thanks to this patch. This wasn't the case before. 

[1] https://github.com/mozilla-releng/shipit#to-rebuild-product-details

### DIFF
--- a/api/src/shipit_api/admin/product_details.py
+++ b/api/src/shipit_api/admin/product_details.py
@@ -345,7 +345,7 @@ def get_releases(
             # product_with_version looks like "Fennec-1.0". There is nothing after the version
             if "-" not in product_with_version:
                 raise ValueError(f'Invalid product_with_version "{product_with_version}". It must contain a -')
-            product_string, version_string = product_with_version.split("-")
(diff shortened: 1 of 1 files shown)


api/src/shipit_api/admin/product_details.py
```diff
@@ -345,7 +345,7 @@ def get_releases(
             # product_with_version looks like "Fennec-1.0". There is nothing after the version
             if "-" not in product_with_version:
                 raise ValueError(f'Invalid product_with_version "{product_with_version}". It must contain a -')
-            product_string, version_string = product_with_version.split("-")
+            product_string, version_string = product_with_version.rsplit("-", 1)
```
JohanLorenzo, line 348: Kudos to @gbrownmozilla for suggesting this solution. I didn't know about this function and it does the job perfectly! 

[1] https://docs.python.org/3.11/library/stdtypes.html#str.rsplit

gbrownmozilla (approved):

PR #1268, opened 2023-01-31
Bug 1813994 -  Remove all deprecated Android products from shipit

JohanLorenzo:
WIP. We have to handle fenix in a couple of weeks.

PR #1295, opened 2023-03-04
Bug 1813994 - Remove deprecated mobile products

gbrownmozilla:
This is just like Johan's https://github.com/mozilla-releng/shipit/pull/1268 (thank you!) but now with Fenix added.

### DIFF
--- a/api/src/shipit_api/admin/api.yml
+++ b/api/src/shipit_api/admin/api.yml
@@ -305,7 +305,6 @@ paths:
           type: string
           enum:
           - firefox
-          - fennec
           - devedition
           - pinebuild
       - name: branch
@@ -720,14 +719,10 @@ components:
           type: string
           example: firefox
           enum:
-          - android-components
           - devedition
           - pinebuild
-          - fenix
-          - fennec
           - firefox
           - firefox-android
-          - focus-android
           - thunderbird
         version:
           type: string
@@ -770,7 +765,7 @@ components:
           example: fennec_beta
         repo_url:
           type: string
-          example: https://github.com/mozilla-mobile/fenix
+          example: https://github.com/mozilla-mobile/firefox-android
     ReleaseStatus:
       required:
       - status
@@ -830,14 +825,10 @@ components:
           type: string
           example: firefox
           enum:
-          - android-components
           - devedition
           - pinebuild
-          - fenix
-          - fennec
           - firefox
           - firefox-android
-          - focus-android
           - thunderbird
         version:
           type: string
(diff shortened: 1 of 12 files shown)


api/src/shipit_api/admin/api.yml
```diff
@@ -770,7 +765,7 @@ components:
           example: fennec_beta
         repo_url:
           type: string
-          example: https://github.com/mozilla-mobile/fenix
+          example: https://github.com/mozilla-mobile/firefox-android
```
JohanLorenzo, line 768: Good idea for updating this example!

api/src/shipit_api/admin/product_details.py
```diff
@@ -191,14 +191,13 @@ async def fetch_l10n_data(
         Product(release.product) is Product.THUNDERBIRD
         and release.branch == "releases/comm-beta"
         and release.revision in ["3e01e0dc6943", "481fea2011e6", "85cb8f907b18", "92950b2fd2dc", "c614b6e7cf58", "e277e3f0ab13", "efd290b55a35", "f87ba53e04ff"]
-    ) or Product(release.product) in (Product.FENIX, Product.ANDROID_COMPONENTS, Product.FOCUS_ANDROID, Product.FIREFOX_ANDROID):
```
JohanLorenzo, line 194: Potential bug: I'd keep all legacy products everywhere in `product_details.py`. I might be wrong but I suspect this removal is going to delete historical data in there[1], for instance. 

[1] https://product-details.mozilla.org/1.0/all.json

api/src/shipit_api/admin/product_details.py
```diff
@@ -191,14 +191,13 @@ async def fetch_l10n_data(
...
+    ) or Product(release.product) == Product.FIREFOX_ANDROID:
         return (release, None)
 
     url_file = {
         Product.FIREFOX: "browser/locales/l10n-changesets.json",
         Product.DEVEDITION: "browser/locales/l10n-changesets.json",
         Product.PINEBUILD: "browser/locales/l10n-changesets.json",
-        Product.FENNEC: "mobile/locales/l10n-changesets.json",
```
JohanLorenzo, line 201: Nice find! Yeah, let's move this line. For sure, we don't store firefox-android locales there.

JohanLorenzo (approved):
Thank you so much for unbiroting my patch! 👍 It looks good to me overall. I'd keep `Product.FENIX`, `Product.ANDROID_COMPONENTS`, and `Product.FOCUS_ANDROID` in `product_details.py` because I suspect it's going to break historical data. More details below.

api/src/shipit_api/admin/product_details.py
gbrownmozilla: Oh, right. Thanks for catching that. I've restored those.

PR #1306, opened 2023-03-23
Bug 1813994 - Cleanup deprecated mobile products

gbrownmozilla:
A more conservative replacement for #1295

### DIFF
--- a/api/src/shipit_api/admin/release.py
+++ b/api/src/shipit_api/admin/release.py
@@ -37,7 +37,11 @@ def parse_version(product, version):
         except KeyError:
             raise ValueError(f"Product {product} versions are not supported")
 
-    VersionClass = _VERSION_CLASS_PER_PRODUCT[product_enum]
+    try:
+        VersionClass = _VERSION_CLASS_PER_PRODUCT[product_enum]
+    except KeyError:
+        raise ValueError(f"Product {product} versions are not supported")
+
     return VersionClass.parse(version)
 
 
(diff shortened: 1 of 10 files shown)


api/src/shipit_api/admin/release.py
```diff
@@ -37,7 +37,11 @@ def parse_version(product, version):
         except KeyError:
             raise ValueError(f"Product {product} versions are not supported")
 
-    VersionClass = _VERSION_CLASS_PER_PRODUCT[product_enum]
+    try:
```
gbrownmozilla, line 40: Left this in, copied from #1295: even though nothing else is changing here, this seems like a good idea.

api/src/shipit_api/admin/settings.py
```diff
@@ -75,11 +75,8 @@
 LDAP_GROUPS = {
     "admin": ADMIN_LDAP_GROUP,
     "firefox-signoff": ["shipit_firefox"],
-    "fenix-signoff": ["shipit_mobile"],
```
gbrownmozilla, line 78: Should be no need to signoff on old products.

api/src/shipit_api/public/api.yml
JohanLorenzo, line 1: Question: I see we still allow deprecated products in the `Release` scheme, unlike what we did in the original PR[1]. What kind of issues this part of the original PR (would have) caused? 

[1] https://github.com/mozilla-releng/shipit/pull/1295/files#diff-872926f584b15572d240a25ce24bf884b1cd0cb584bac73020b7c4cf4a47e390

JohanLorenzo (approved):
Looks good to me! I agree with this more conservative approach. I look forward to seeing this cleanup in action 😃 

I mentally diff'd this current PR with the original one (https://github.com/mozilla-releng/shipit/pull/1295). If I understand correctly and for the record, they're the same expect these files didn't get changed:

 * api/src/shipit_api/admin/release.py 
 * api/src/shipit_api/public/api.py
 * api/tests/test_release.py

gabrielBusta (approved):

PR #1316, opened 2023-05-03
feat: Add birch branch to dev config

hneiva:


frontend/src/configs/dev.js
```diff
@@ -61,6 +61,22 @@ module.exports = {
       enablePartials: true,
       canTogglePartials: true,
     },
+    {
+      product: 'birch',
+      prettyName: 'Firefox birch',
+      appName: 'browser',
+      branches: [
+        {
+          prettyName: 'Try',
```
bhearsum, line 70: These three strings should be talking about birch - not try.

bhearsum: It also looks like the branch should be `projects/birch` based on how I see Beta configure in production: https://github.com/mozilla-releng/shipit/blob/172d396f61bb9873b9ebc1ab2214a54d6b8c07f0/frontend/src/configs/production.js#L19. (It seems that we need the path part of the url here.)

bhearsum (changes requested):
I'd suggest you add this to `development` as well just to be safe. Or you can try this one, and then switch to that one if needed.

bhearsum (approved):

jcristau (commented):
Can you explain why we need/want this?

bhearsum:
> Can you explain why we need/want this?

I recommended that we do this in order to make it possible to test partner & eme-free notarization tasks. (Something we can't do on Try AFAIK.)

jcristau:
We can't test notarization on try but we don't need shipit either, we can trigger release-promotion graphs from th or tc directly?

bhearsum:
> We can't test notarization on try but we don't need shipit either, we can trigger release-promotion graphs from th or tc directly?

It's not strictly necessary, yeah, but it's a bit more convenient? I could be convinced not to bother with this...

gbrownmozilla (approved):

jcristau (changes requested):
One thing is I'm a bit worried about mixing a level3 repo with staging infrastructure.

Relatedly, and probably more importantly, the project/releng/shipit/staging tc client only has scopes to trigger level1 hooks.

bhearsum:
In any case - I think we _really_ ought to have a way to test notarization changes outside of landing them to central - at least in the short term while this is under active development.

bhearsum:
@jcristau and I chatted about this elsewhere. I got a bit mixed up in my previous comment - we can still test release promotion without this by kicking it off directly with hooks instead of Ship It (and it looks like we'll have to due to ship it scopes).

hneiva:
Sounds like it's safe to close this for now, as it's not needed.

PR #1317, opened 2023-05-08
[do not merge until May 9] 115 version bumps

bhearsum:


api/src/shipit_api/common/config.py
```diff
@@ -49,17 +49,17 @@
...
-# They can be found: https://wiki.mozilla.org/Release_Management/Calendar
-LAST_SOFTFREEZE_DATE = "2023-04-06"
-LAST_MERGE_DATE = "2023-04-10"
-LAST_RELEASE_DATE = "2023-04-11"
-NEXT_SOFTFREEZE_DATE = "2023-05-04"
-NEXT_MERGE_DATE = "2023-05-08"
-NEXT_RELEASE_DATE = "2023-05-09"
+# They can be found on https://whattrainisitnow.com/calendar/ 
+LAST_SOFTFREEZE_DATE = "2023-05-04"
+LAST_MERGE_DATE = "2023-05-88"
```
gbrownmozilla, line 58: Oops! "88"!?

gbrownmozilla (changes requested):

api/src/shipit_api/common/config.py
bhearsum: Ermmmmm, yes.

gbrownmozilla (approved):

Issue #1323, opened 2023-05-31
IOS_VERSION has not been updated in 3 years


gabrielBusta:
`IOS_VERSION` has not been updated in very long. It is very stale and it is unlikely anyone is relaying on it. We are on [v113.2](https://github.com/mozilla-mobile/firefox-ios/releases/tag/v113.2) on iOS. We should return an "" like we do on other deprecated fields.

JohanLorenzo:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1949689

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #1324, opened 2023-05-31
Update config.py to return an "" for iOS version

gabrielBusta:
I found this while reading through https://bugzilla.mozilla.org/show_bug.cgi?id=1643121


gabrielBusta:
#1323

PR #1329, opened 2023-06-27
Bug 1835641 - add esr-102 as alternative to esr-115

gbrownmozilla:
I think this was missed in #1328 .


jcristau (commented):
comm-esr115 also seems to be missing?

jcristau (approved):

jcristau:
(adding comm-esr115 in #1330)

PR #1336, opened 2023-07-14
Update the Ship-It navigation to fit more groups

gabrielBusta:
Updates the shipit navigation to show a releases menu that re-uses the existing shipit navigation.

### DIFF
--- a/frontend/src/Main.jsx
+++ b/frontend/src/Main.jsx
@@ -33,7 +33,7 @@ function Main() {
   if (backendStatus.loading) {
     return (
       <BrowserRouter>
-        <Dashboard title="Loading" disabled>
+        <Dashboard disabled>
           <Spinner loading />
         </Dashboard>
       </BrowserRouter>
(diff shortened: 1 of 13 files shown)


gabrielBusta:
Thanks @ahal

> Make the RELEASES menu item a little bigger and [...] include a down arrow [...] to indicate that it is a drop down

I matched the menu button's typography to the rest of the navigation and added a drop down arrow.

> you could include the name of the product to the left of the navbar

I added breadcrumbs to the navbar.

> When clicking a product to uncollapse, maybe it could automatically collapse any previously clicked item

I like the idea, but this item is a little more difficult because I'd have to add some state to the app. Let's leave that as a follow up. I am sure it can be done.

gbrownmozilla (approved):
I like it! Thanks.

One thing that bothered me a little was "Settings"; do you think we will add anything here? If not, it seems odd to have two product details actions there. Maybe "Product Details" > "Update" / "View" would work instead?

gabrielBusta:
I have changes based on this, I will try in a separate PR.

PR #1338, opened 2023-07-26
App services role

ahal:


api/src/shipit_api/admin/settings.py
```diff
@@ -74,6 +74,7 @@
 
 LDAP_GROUPS = {
     "admin": ADMIN_LDAP_GROUP,
+    "relman": ["shipit_relman"],
```
ahal, line 77: TODO: create this group.

ahal:
I think this is ready for a first pass review, though not landing yet.

api/src/shipit_api/admin/settings.py
```diff
@@ -93,20 +95,20 @@
 
 AUTH0_AUTH_SCOPES = dict()
 
-# releng signoff scopes
+# firefox scopes
+firefox_ldap_groups = sorted(set(LDAP_GROUPS["firefox-signoff"] + LDAP_GROUPS["relman"]))
```
gbrownmozilla, line 99: I'm slightly surprised to see relman added for firefox here -- is it kind-of a ride-along enhancement? What's the current situation for relman people -- are they all in firefox-signoff?

ahal: Yeah, I guess it's a bit of a ride along and I could drop it from this changeset. And yes, afaict everyone in relman has both `shipit_firefox` and `shipit_mobile` (so adding the `relman_shipit` group here would be a no-op in practice).

My reasoning for doing this here is:
1. Keep things consistent  across Firefox, app-services and future shipit projects (having some projects rely on a `relman` group and others not seems like it would be confusing).
2. Make less work for IT (new relman members only need to be added to one group rather than N)
3. Leave the possibility open for dual signoff if Firefox products (not sure we'll ever do this due to dual signoff already happening in Balrog.. but maybe product teams would appreciate it in some circumstances?)

gbrownmozilla: That makes sense. Thanks for the explanation.

gbrownmozilla (approved):

gabrielBusta (approved):

gabrielBusta (approved):

ahal:
This depends on #1351 landing first.

PR #1339, opened 2023-07-27
Use protocol (Mozilla's design system)

gabrielBusta:
Use Mozilla's design system in Ship-It.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -34,6 +34,7 @@
     "@material-ui/pickers": "3.3.10",
     "@material-ui/styles": "4.11.5",
     "@mozilla-frontend-infra/components": "3.0.7",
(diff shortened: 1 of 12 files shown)


gabrielBusta:
Its in staging.

gabrielBusta:
I will follow up on

> "Product Details" > "Update" / "View

from #1336 in another PR when I get a chance @gbrownmozilla

I wanna re-use the drop-down on the releases menu to get that done.

> do you think we will add anything here?

Yeah, I'd like to add something like "Create Product" (that would be some work), maybe a way to update phase sign-off sign-offs/permissions (like in balrog), or maybe a UI to update the product detail versions on a table instead of having them hard-coded, just some thoughts.

gbrownmozilla (approved):

PR #1345, opened 2023-08-01
Print a bunch of things about the GitHub graphql response

bhearsum:


api/src/shipit_api/admin/github.py
```diff
@@ -87,6 +87,11 @@ def query_api(query):
     req = requests.post(GITHUB_API_ENDPOINT, json={"query": query}, headers=headers)
     req.raise_for_status()
 
+    print(req)
+    print(req.headers)
+    print(req.status_code)
+    print(req.text)
```
JohanLorenzo, line 93: Nit: I'd remove `req.text` since `req.content` already takes care of printing the raw body https://docs.python-requests.org/en/latest/user/quickstart/#binary-response-content

api/src/shipit_api/admin/github.py
```diff
@@ -87,6 +87,11 @@ def query_api(query):
     req = requests.post(GITHUB_API_ENDPOINT, json={"query": query}, headers=headers)
     req.raise_for_status()
 
+    print(req)
```
JohanLorenzo, line 90: Nit: Let's use `logger.error()` just  in case `print()` gets ignored. It's already used in that file at:

https://github.com/mozilla-releng/shipit/blob/91fefdf09c84c157015c73d794a247582c0095a9/api/src/shipit_api/admin/github.py#L201

JohanLorenzo (approved):
Looks good to me modulo a couple nits.

api/src/shipit_api/admin/github.py
bhearsum: I'm going to do both. I don't know offhand what the log level is set to, and I can't find any examples of this message being printed already.

PR #1350, opened 2023-08-01
Update shipit api dependencies and fix Github gzip issue

ahal:
I figured it wouldn't hurt to try and figure out that Ship-It gzip issue sooner rather than later (before we all forget about it). Unfortunately I don't seem to be able to reproduce the issue on a local instance, so I'm going to try and reproduce on the staging instance.


ahal:
Ok so I pushed this to `dev`, but I can't reproduce the error on staging shipit.. so I'm not sure what to do.. I guess we could land it on prod?

PR #1353, opened 2023-08-03
Enable phase skipping for app-services

ahal:
This allows us to configure where phase skipping is allowed in `config.py` and enables it for app-services.

### DIFF
--- a/api/src/shipit_api/common/config.py
+++ b/api/src/shipit_api/common/config.py
@@ -544,3 +544,14 @@ SIGNOFFS = {
         },
     },
 }
+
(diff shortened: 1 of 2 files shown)


ahal:
IIRC we used to always be able to skip phases for all products. Should we revisit turning it back on?

ahal:
> IIRC we used to always be able to skip phases for all products. Should we revisit turning it back on in more places?

I guess I was mistaken and it *was* only beta / try where this was possible.

gabrielBusta (approved):

PR #1358, opened 2023-08-28
Update current and next ESR versions

ahal:
I missed this in the regular product details update as most cycles we don't decomission an ESR.


jcristau (changes requested):
This should only land in 4 weeks, AFAIK.

ahal:
@jcristau Ah ok, thanks for catching! Do you know why that is?

gbrownmozilla (approved):

jcristau (approved):
Should land/deploy today after the 115.3.0 release.

PR #1364, opened 2023-10-04
Enable firefox-android release promotion on try (bug 1825123)

jcristau:
- add new phases to the API for the firefox-android product to match what the new tasks are called
- add the try repo as an option for firefox-android builds on dev and development environments

### DIFF
--- a/frontend/src/configs/dev.js
+++ b/frontend/src/configs/dev.js
@@ -69,6 +69,14 @@ module.exports = {
         {
           branch: '',
         },
+        {
+          prettyName: 'Try',
+          project: 'try',
+          branch: 'try',
+          repo: 'https://hg.mozilla.org/try',
+          enableReleaseEta: false,
+          versionFile: 'mobile/android/version.txt',
+        },
       ],
       repositories: [
         {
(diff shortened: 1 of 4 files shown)


frontend/src/configs/dev.js
```diff
@@ -69,6 +69,14 @@ module.exports = {
         {
           branch: '',
```
gbrownmozilla, line 70: Do we still need this entry, or was it just a placeholder?

gbrownmozilla (approved):

frontend/src/configs/dev.js
jcristau: I *think* we still need it while github is around, but I'll double check.

gabrielBusta (approved):

frontend/src/configs/dev.js
jcristau: @gbrownmozilla thanks for making me double check this, it was broken.

It looks like in the config we either list "branches", which actually means hg repos, or we list "repositories", which means github.
I think I got things to work by making it possible to list both a github repo and a hg "branch" for the same product in the config, whereby the "new release" ui uses the "branches" from the config file if no repository is selected, but if a repository is selected it uses the list of branches queried from github.

jcristau: There's still bits I'm unsure about though, e.g. https://github.com/mozilla-releng/shipit/blob/c700e2d3be88e61f334752ed201fc437f90cbaab/frontend/src/views/Releases/index.jsx#L40-L41
so I'll go back and test some more...

jcristau: Right, so if you don't keep the empty branch in the config for that product, then the "recent releases" view only queries for `product=firefox-android&branch=try`, and so it won't show the releases from github.

PR #1365, opened 2023-10-10
Dependency upgrades

bhearsum:
Two notable things:
- `category` getting dropped from the lockfile (I believe this is expected, see https://github.com/python-poetry/poetry/issues/5702)
- `werkzeug` is being downgraded

I believe both of these things are caused by me running with Poetry 1.6.1, and no big deal.


jcristau:
@bhearsum we're pinning poetry to 1.4.2 in api/requirements/poetry.txt, should that get updated too?

bhearsum:
> @bhearsum we're pinning poetry to 1.4.2 in api/requirements/poetry.txt, should that get updated too?

Ah, indeed! Yes - may as well.

ahal (approved):

PR #1372, opened 2023-11-07
chore: bootstrap Poetry in CI via pipx

ahal:
The poetry installation docs state very clearly that you can't install Poetry into the same Python environment as the one it is management. The old method of `pip install -r poetry.txt` was doing exactly that.

Instead, let's use `pipx` which hides Poetry away in an isolated environment. While this doesn't check hashes, it's likely not a big deal for this case.

As far as reproducing locally, I recommend also using `pipx`. It's even possible to use `pipx` to install multiple versions of Poetry:

    $ pipx install --suffix shipit poetry==<version>
    $ poetry@shipit install

This way, if you *really* need the exact same version that Shipit CI uses, it's possible.


ahal:
Not sure what's going on here, but I'm unlikely to find time to dig into it for awhile.

PR #1382, opened 2023-12-19
 chore(api): update dependencies

gbrownmozilla:


bhearsum (approved):
Note: we should avoid deploying this to prod before the new year due to the ongoing change freeze.

PR #1384, opened 2023-12-20
Add mozilla-vpn-client and addons to shipit

ahal:
This also allows the release views (new, pending, releases) to take a "group" via url parameter. Only repositories configured under said group will be shown. The `PRODUCTS` key in the configuration is refactored such that projects are listed under the specified group. By default, "firefox" is used if no group was specified via param.

This way we can add new groups to the dropdown without needing separate views for each one.

### DIFF
--- a/api/src/shipit_api/admin/api.yml
+++ b/api/src/shipit_api/admin/api.yml
@@ -722,6 +722,8 @@ components:
           - devedition
           - firefox
           - firefox-android
+          - mozilla-vpn-addons
+          - mozilla-vpn-client
           - thunderbird
         version:
           type: string
@@ -832,6 +834,8 @@ components:
           - firefox
           - firefox-android
           - focus-android
+          - mozilla-vpn-addons
+          - mozilla-vpn-client
           - thunderbird
         version:
           type: string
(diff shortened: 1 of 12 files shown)


ahal:
I believe this is ready for review, things appear to work more or less as I'd expect them to. This change is live on the dev instance, so you can play around with it over there.

The only "bug" is that if you manually type in an invalid group and get the error banner, switching to a valid group doesn't automatically re-render and a refresh is needed. I couldn't figure out how to fix this, but people shouldn't be manually typing in groups anyway so this isn't a big deal imo.

The other thing that we might want to change is that the "group" should likely be part of the path rather than a URL parameter. But I don't understand React (pretty much at all) and since I managed to get this working, I don't really have the time or motivation to figure out how to make that switch. If anyone else wants to make this change, please feel free! But otherwise the current implementation feels good enough to me.

ahal:
I found another bug:

When you trigger a promotion phase, the releases disappear and do not re-render. I'm not sure what is causing this but will investigate.

ahal:
I think I'm also going to split the VPN stuff to a separate PR and keep this one focused on adding groups.

ahal:
Actually I spun the groups PR off into:
https://github.com/mozilla-releng/shipit/pull/1385

So I'm going to put this one on hold until we can get that merged.

ahal:
We should block on changes to the VPN repo before we start using Shipit there, but I see no harm in getting this reviewed + landed in the meantime.

jcristau (approved):
Looks sane to me from a cursory review.

PR #1385, opened 2024-01-04
Support grouping releases by high level product areas

ahal:
Currently releases can either be under "Firefox" or "XPI", the latter of which implements entirely new views with separate logic.

In order to support non-Firefox related products (like VPN), we should have the ability to display releases only related to a top-level group. For example, when you click "Pending" under Firefox, you should see releases related to all Firefox repos, but not VPN ones.

If you click "Pending" under a theoretical new "Security" grouping, you should only see releases related to VPN and not Firefox.

This way, you get a holistic view of all releases that are related to one another without having the UI cluttered with unrelated releases.


ahal:
This is mostly working except for one issue:

If you create a release then trigger one of the phases, the pending releases don't get re-rendered. Instead you get a blank page. If you refresh (either in-browser or via the button) the pending releases show up again.

I'm not sure why my changes caused this behaviour. I've spent some time digging into it but am so far at a loss.

gabrielBusta:
The issue is that when the UI calls the "release fetcher" function, it is doing it without passing in any parameters.

https://github.com/mozilla-releng/shipit/blob/fef5858389dbbb71741c563979e9782758d99cb6/frontend/src/components/PhaseProgress/index.jsx#L69

The "release fetcher" is expecting a `productBranches`

https://github.com/mozilla-releng/shipit/blob/8282986dd9400b8ab3f9df1406ab7eeca74c8069/frontend/src/views/Releases/index.jsx#L46

I fixed the issue on https://github.com/mozilla-releng/shipit/commit/b751dab14cc3ef0c0e471dfb5e9be6e5e51d3890 by saving the `productBranches` in that context object we are using  to store the reference to `fetchReleases`. That might not be the best way to fix this. I didn't get a chance to try it when there's a different value for the release fetcher. They may not all take in this argument, not sure if that is an issue.

ahal:
Thank you so much! I wonder if an alternative solution could be to move the `getProductBranches` call into the `fetchReleases` function. I think I tried that and hit some problem, but maybe I'll give it another go. Would be nice if we could just pass that one function around that doesn't need any args.

ahal:
So the above works, but now I seem to have broken creating XPI releases.. I think this issue will be a bit easier to debug though.

frontend/src/views/Releases/index.jsx
```diff
@@ -15,7 +18,21 @@ import ReleaseContext from '../../utils/ReleaseContext';
...
+  return config.PRODUCTS[group]
+    .map(p => [p.branches.map(b => [p.product, b.branch])])
+    .flatMap(x => x)
+    .flatMap(x => x);
+}
+
 export default function Releases({ recent = false, xpi = false }) {
+  const location = useLocation();
+  const group = new URLSearchParams(location.search).get('group') || 'firefox';
+  const groupTitle = group.charAt(0).toUpperCase() + group.slice(1);
```
gbrownmozilla, line 35: nit: I dislike having a difference between what's shown in the UI and what's in the data. I think I wouldn't mind seeing "firefox" displayed.

frontend/src/configs/production.js
```diff
@@ -9,145 +9,147 @@ module.exports = {
...
-          repo: 'https://github.com/mozilla/application-services',
-          enableReleaseEta: false,
-          enableTreeherder: false,
-        },
-      ],
-      enablePartials: false,
-    },
-  ],
+  PRODUCTS: {
+    firefox: [
```
gbrownmozilla, line 13: nit: It seems odd having thunderbird and app-services under "firefox". I wonder if there's a better name. ...but I don't like "gecko", and I'm not sure I like "default". (Maybe "firefox" is the best we can do.)

gbrownmozilla (approved):
This is awesome - thanks!

frontend/src/configs/production.js
ahal: Yeah, I mostly went with "Firefox" because that's what it's already listed as in the menu / header. Maybe "Firefox" isn't the best name, but I do think they all belong together.

App-services feeds into Firefox, and their release blocks the main Firefox one. So I think it makes sense that it would show up in the same view as the regular Firefox release. And Thunderbird has the same release cycle, so I think that sort of makes sense.. though we could always make a dedicated "Thunderbird" group if we wanted.

gabrielBusta: Maybe we could group them under Gecko? Is that what they have in common?

gabrielBusta: I don't like gecko that much either tho

ahal: Yeah, I don't really care too much, I'd be fine leaving as is or switching to `Gecko` or creating a new `Thunderbird` group. Either way, it should be a follow-up (not least because I already landed this :p)

PR #1387, opened 2024-01-10
fix: redirect to the pending releases page after submitting a release

### LINKED ISSUE #1355: ship it should redirect to "pending releases" after submitting one
Right now, it simply refreshes the new release page - which is almost never what you want. Typically the next action taken is to start a phase - which happens on the Pending Releases page.

gabrielBusta:
fixes #1355

### DIFF
--- a/frontend/src/views/NewRelease/index.jsx
+++ b/frontend/src/views/NewRelease/index.jsx
@@ -1,6 +1,6 @@
 import 'date-fns';
 import React, { useState, useContext } from 'react';
-import { useLocation } from 'react-router-dom';
+import { useLocation, useHistory } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


gabrielBusta:
Should I be passing in the "group" somewhere when redirecting back to the releases page?

ahal:
> Should I be passing in the "group" somewhere when redirecting back to the releases page?

Yeah, it should be a url parameter, e.g: `/?group=security`. It'll default to "firefox" if unspecified, so no need to pass it if `group == "firefox"` (but it doesn't hurt to pass it either.

ahal (changes requested):
Requesting changes for the 'group' issue

ahal (approved):

bhearsum (approved):

PR #1397, opened 2024-02-07
Bug 1879094 - part 7 to 11: Centralize some of the product definitions in `api/products.yml`

JohanLorenzo:
Test coverage could be increase here. I'd like to finish centralizing the definitions first before adding the missing coverage. How does that sound?


gabrielBusta:
Nice!

bhearsum (approved):

PR #1398, opened 2024-02-12
Bug 1879094 - part 12 to 23: Finish centralizing the product config in the backend

JohanLorenzo:
Depends on https://github.com/mozilla-releng/shipit/pull/1395 and https://github.com/mozilla-releng/shipit/pull/1397


api/products.yml
```diff
@@ -3,16 +3,27 @@ android-components:
   legacy: true
   version-class: "mozilla_version.mobile.MobileVersion"
 app-services:
+  ldap-groups:
```
bhearsum, line 6: suggestion (non-blocking): Something like `authorized-ldap-groups` would be more explicit that this is the set of groups that can create releases for this app. (I feel like there's probably an even better name for this, but I'm struggling to find it...)

api/products.yml
```diff
@@ -3,16 +3,27 @@ android-components:
...
+  - shipit_relman
   # Phases per product, ordered
   phases:
   - promote
   - ship
+  firefox-ci-trust-domain: app-services
+  repo-url: https://github.com/mozilla/application-services
   version-class: "mozilla_version.gecko.GeckoVersion"
 devedition:
+  can-be-disabled: true
```
bhearsum, line 18: nit: `disablable` reads slightly nicer.

api/src/shipit_api/admin/settings.py
```diff
@@ -96,88 +98,64 @@
...
+    ldap_groups_per_scope = {}
+    for product_name, product_config in get_products_config().items():
+        if product_config["legacy"]:
+            continue
+
+        scopes = _get_auth0_scopes(product_name, product_config)
+        new_ldap_groups_per_scope = {s: product_config["ldap-groups"] for s in scopes}
+        ldap_groups_per_scope = merge_or_raise.merge(ldap_groups_per_scope, new_ldap_groups_per_scope)
+
+    return merge_or_raise.merge(AUTH0_AUTH_SCOPES, ldap_groups_per_scope)
```
bhearsum, line 112: question: What's the purpose of using `merge_or_raise` here? `AUTH0_AUTH_SCOPES` is always an empty dict, and we overwrite it with the return from this function further down. It seems like we should just be able to return `ldap_group_per_scope` directly here.

api/src/shipit_api/admin/settings.py
```diff
@@ -96,88 +98,64 @@
...
-        [
-            f"add_release/{flavor}",
-            f"abandon_release/{flavor}",
-        ]
-    )
-    phases = [i["name"] for i in SUPPORTED_FLAVORS.get(flavor, [])]
-    for phase in set(phases):
-        scopes.extend([f"schedule_phase/{flavor}/{phase}", f"phase_signoff/{flavor}/{phase}"])
-AUTH0_AUTH_SCOPES.update({s: LDAP_GROUPS["vpn-signoff"] for s in scopes})
+def _assign_ldap_groups_to_scopes():
```
bhearsum, line 102: suggestion (non-blocking): Consider putting these new helper functions in a utility file. The more code that exists here, the more difficult it is to determine what the settings are at a glance.

bhearsum (approved):
I'm marking this approved but I have a few questions and non-blocking suggestions. This is great clean up!

api/products.yml
JohanLorenzo: Fun fact: I asked @ahal and @hneiva about this very word last week. I couldn't find a dictionary defining this word. So we came up with `can-be-disabled` as consensus

hneiva: Hardest problem in computer science!

PR #1400, opened 2024-02-19
Bug 1880524 - revert changes from bug 1879094

jcristau:
This reverts the changes from bug 1879094 for introducing a regression.


jcristau:
The regression was fixed by #1402 instead, closing.

PR #1406, opened 2024-02-21
admin: find trust domain based on repo rather than product config

### LINKED ISSUE #1405: `IndexError: list index out of range` when creating a firefox-android staging release from try
Sentry Issue: [SHIPIT-API-3A](https://mozilla.sentry.io/issues/4999188555/?referrer=github_integration)

```
IndexError: list index out of range
(9 additional frame(s) were not displayed)
...
  File "/app/src/src/shipit_api/admin/api.py", line 114, in add_release
    release.phases = generate_phases(release, common_input, verify_supported_flavors=True)
  File "/app/src/src/shipit_api/admin/tasks.py", line 201, in generate_phases
    decision_task_id = find_decision_task_id(release.repo_url, release.project, release.revision, release.product)
  File "/app/src/src/shipit_api/admin/tasks.py", line 50, in find_decision_task_id
    _, project = extract_github_repo_owner_and_name(repo_url)
  File "/app/src/src/shipit_api/admin/github.py", line 282, in extract_github_repo_owner_and_name
    repo_name = parts[1]
```

jcristau:
This fixes a regression where we would treat try as a github repo when creating a firefox-android staging release, because we assumed it should use the "mobile" trust domain.

Closes #1405

### DIFF
--- a/api/tests/test_tasks.py
+++ b/api/tests/test_tasks.py
@@ -114,3 +114,24 @@ def test_extract_our_flavors_warnings(caplog, avail_flavors, expected_records, e
     if expected_records > 0:
         assert caplog.records[0].levelname == "WARNING"
         assert expected_text in caplog.text
+
+
+@pytest.mark.parametrize(
+    "repo_url, product, expectation, trust_domain",
+    (
+        ("https://hg.mozilla.org/try", "firefox-android", does_not_raise(), "gecko"),
+        ("https://hg.mozilla.org/try", "firefox", does_not_raise(), "gecko"),
+        ("https://hg.mozilla.org/releases/mozilla-beta", "firefox-android", does_not_raise(), "gecko"),
+        ("https://hg.mozilla.org/mozilla-central", "firefox-android", pytest.raises(KeyError), None),
+        ("https://hg.mozilla.org/try-comm-central", "firefox-android", does_not_raise(), "comm"),
+        ("https://hg.mozilla.org/try-comm-central", "thunderbird", does_not_raise(), "comm"),
+        ("", "xpi", does_not_raise(), "xpi"),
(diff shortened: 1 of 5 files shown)


api/tests/test_tasks.py
```diff
@@ -114,3 +114,24 @@ def test_extract_our_flavors_warnings(caplog, avail_flavors, expected_records, e
     if expected_records > 0:
         assert caplog.records[0].levelname == "WARNING"
         assert expected_text in caplog.text
+
+
+@pytest.mark.parametrize(
```
gbrownmozilla, line 119: Thanks for tests!

gbrownmozilla (approved):

JohanLorenzo (approved):
Thank you very much for following up on the big refactor I made! 

Let's get this deployed so progress can resume on the Android repo migration project. 

In the long run, I'm currently not convinced we should set `trust-domains` based on the repo URL. Looking at `trust-domains.yml`, the only exception to trust-domain not being directly tied to the product is `firefox-android`. So we're driving this change because of one exception. 
In addition to that, the main point of the refactor was to centralize the configuration into a single file. Creating `trust-domains.yml` feels backwards to me.

I'm probably missing something, though. I'm happy to discuss this further 🙂

jcristau:
The trust domain is fundamentally tied to the repo, not to the product.  Yes usually that won't make a difference, but the fact that there's currently an exception is evidence that the change to tie it to the product was flawed.

I also think "configuration should be in a single file" is fairly weak of an argument.  I understand the desire to move configuration to something declarative rather than python code, so this change doesn't backtrack on that.  But if we want a single file we can e.g. have a config.yml with products and trust-domains keys to replace products.yml and trust-domains.yml.

PR #1414, opened 2024-03-18
Bug 1859885 - Temporarily use Beta for comm-release partials

jfx2006:
Handle the proverbial chicken and egg scenario. Partial updates for comm-release cannot be generated as there's no prior releases in Shipit's database to use as a base. Turning on "canTogglePartials" for Thunderbird won't help since Taskgraph expects to receive a "partial_updates" field in the promote, push, and ship payloads. That's not easy to change due to the config.yml schema.

For now, use comm-beta as an alternate repository for partials, like for new esr repositories. Once a few release candidates have shipped this can be removed.


gabrielBusta (approved):

gabrielBusta:
🤞

PR #1416, opened 2024-03-25
Bug 1825123 - enable firefox-android releases on mozilla-beta and mozilla-release

jcristau:


frontend/src/configs/production.js
```diff
@@ -83,6 +83,24 @@ module.exports = {
           {
             branch: '',
```
gbrownmozilla, line 84: Is this "empty" branch still needed, once Beta/Release are added? What does it do?

jcristau: It's needed until we're entirely off github.

jcristau: (i.e., we should keep both for one cycle, and we can remove the empty branch once 126 is shipped to release and 125 is EOL)

gbrownmozilla (approved):

PR #1417, opened 2024-03-26
Update Python dependencies

bhearsum:
Only major bump is to `packaging`, for:
> BREAKING: Make optional metadata.Metadata attributes default to None (733)

Assuming tests pass here, that should be a good enough sanity check.


gabrielBusta (approved):

bhearsum:
> Only nit is that you used an older version of poetry to generate this than what was used last time.. but I doubt it makes a difference.

oops - I thought I updated Poetry first. It looks like you're right that it didn't matter in this case though:
```
~/repos/shipit/api deps5 *10 ❯ poetry --version                                        5s Node system task 4
Poetry (version 1.8.2)
~/repos/shipit/api deps5 *10 ❯ poetry update                                              Node system task 4
Updating dependencies
Resolving dependencies... (2.0s)

No dependencies to install or update

```

PR #1420, opened 2024-04-08
Mount ./api/products.yml to /app/products.yml in development container

gabrielBusta:
If we add this mount the file will change in the container when it is changed in the local checkout, instead of being statically COPY'd into the image.


gbrownmozilla (approved):
Does the quoting work? Is it needed?

gabrielBusta:
the . (dot) is a special character [(e.g.)](https://github.com/mozilla-releng/shipit/blob/main/api/products.yml#L15)

PR #1427, opened 2024-04-16
Mount ./api/trust-domains.yml to /app/trust-domains.yml in development container

gabrielBusta:
So we don't have to delete/rebuild the image locally for the container to pick up changes to trust-domains.yml


bhearsum (approved):
good idea!

gbrownmozilla (approved):

PR #1429, opened 2024-04-17
Bug 1891921 - part 2: Use mozilla-version to determine the category o…

JohanLorenzo:
…f a release

Depends on https://github.com/mozilla-releng/mozilla-version/pull/109


JohanLorenzo:
See what the next commit in product details will look like: https://gist.github.com/JohanLorenzo/744122b093f6d5edfbe46488f91aedc0. 

There are several type of changes: 
* Missing [Fenix beta versions are added](https://gist.github.com/JohanLorenzo/744122b093f6d5edfbe46488f91aedc0#file-product-details-patch-L15). That's because the scheme we used in beta didn't follow [this regex](https://github.com/mozilla-releng/shipit/pull/1429/files#diff-60e8b2b117044af9d445865c8fe0eaada09f61ac1353e8ecf6779bf77616cb11L285)
* [ESRs are correctly categorized](https://gist.github.com/JohanLorenzo/744122b093f6d5edfbe46488f91aedc0#file-product-details-patch-L180) as such (instead of being major/stability) ones. I'm not sure why we ever [categorized them as major/stability](https://github.com/mozilla-releng/shipit/pull/1429/files#diff-60e8b2b117044af9d445865c8fe0eaada09f61ac1353e8ecf6779bf77616cb11L281) while we have the [ESR category too](https://github.com/mozilla-releng/shipit/pull/1429/files#diff-995caf51f06540cfaa44d920df80a2db8a2a03ff868a4c9fa8f4fc6579416507L11). A direct consequence of that is [the disappearance of ESRs](https://gist.github.com/JohanLorenzo/744122b093f6d5edfbe46488f91aedc0#file-product-details-patch-L541) (versioned as X.Y.Z) from the major/stability json files.
* Edge cases [like 14.0.1 are correctly categorized](https://gist.github.com/JohanLorenzo/744122b093f6d5edfbe46488f91aedc0#file-product-details-patch-L530) as major

rvandermeulen:
To be honest, I'm not sure how to even evaluate what the downstream implications of these changes are. I would definitely like @pascalchevrel to take a look when he's back next week as I suspect he may have some better knowledge about who the downstream consumers are of these json files.

As a I mentioned to jlorenzo already, I'd definitely like to confirm that this won't cause the ESR point releases to disappear from https://www.mozilla.org/firefox/releases/ to start.

pascalchevrel:
in the patch, in  `firefox_history_stability_releases.json`:

```patch
-    "125.0.2": "2024-04-22"
+    "125.0.2": "2024-04-16"
```
the initial date of "2024-04-22" was correct

Apart from that, I don't see changes that would obviously break stuff that I know of. I tested the patched version of the files with whattrainisnitnow.com which makes heavy use of product-details data and it works. LGTM

JohanLorenzo:
Status update: I plan to update my patch to avoid making too many changes to product-details. I chatted with the Bedrock team and it's better if we keep the data the way it is. Otherwise, we will break https://www.mozilla.org/en-US/firefox/releases/, for instance. I'll get back to this PR when I'm back from vacation.

gbrownmozilla (changes requested):
Requesting changes based on the April 30 comment.

PR #1434, opened 2024-04-17
better hack to get 125.0.1 in product details rebuilds

bhearsum:


gabrielBusta (approved):

api/src/shipit_api/admin/product_details.py
```diff
@@ -454,6 +454,23 @@ def get_release_history(
         if release_version is None or release_version.major_number < breakpoint_version:
             continue
 
+        # short term hack: 125.0.1 is a major release. we should replace this with
+        # something that uses MozillaVersion to determine categories
+        if (
```
gabrielBusta, line 459: It's an edge case

gabrielBusta (approved):

PR #1439, opened 2024-04-23
chore: update dependabot.yml

ahal:


.github/dependabot.yml
```diff
@@ -1,18 +1,39 @@
+---
 version: 2
 updates:
 - package-ecosystem: pip
-  directory: "/api/requirements"
+  directory: "/api"
```
ahal, line 5: Previously this was only updating the `/api/requirements/poetry.txt` file, which is in turn only used by the Docker image to install dependencies. I guess we could keep having dependabot update this dir.. but it doesn't feel terribly important to do so.

.github/dependabot.yml
```diff
@@ -1,18 +1,39 @@
...
+      patterns:
+        - "*"
+      update-types:
+        - "minor"
+        - "patch"
+  commit-message:
+    prefix: "chore(api)"
+  reviewers:
+    - "mozilla-releng/releng"
 - package-ecosystem: npm
```
bhearsum, line 22: I suggest we remove npm for now. We haven't been taking updates here because we have no testing for this :(.

.github/dependabot.yml
```diff
@@ -1,18 +1,39 @@
...
-      update-types: ["version-update:semver-patch"]
+  open-pull-requests-limit: 5
+  groups:
+    deps:
+      applies-to: version-updates
+      patterns:
+        - "*"
+      update-types:
+        - "minor"
+        - "patch"
```
bhearsum, line 17: If I understand correctly, this means that major versions will never be done by dependabot?

Ideally, I think we'd have major bumps come in via their own PRs (or a group of all of the major ones). (Otherwise it seems likely we never do them until something break.)

ahal: My understanding is that this would mean that they wouldn't get grouped, but would still come in via their own PRs.. But tbh I'm not entirely sure! There's going to be some trial and error here

.github/dependabot.yml
ahal: Or should we keep it so we can see the PR and get motivated by how out of date everything is? :p

ahal: I made the following changes:

- changed schedule from weekly -> monthly
- changed pull requests limit from 5 -> 1
- removed `mozilla-releng/releng` from reviewers
- grouped major bumps alongside minor and patch (since there can only be one PR now)

Does that sound like a good compromise? This way we can still see what needs updating. I'll communicate with the team that this PR needs to be manually tested on `dev` before merging.

bhearsum: Yup, sounds good to me!

.github/dependabot.yml
bhearsum: Ah, OK. Let's see what happens!

PR #1450, opened 2024-04-26
chore(api): remove transient 'itsdangerous' dependency from pyproject…

ahal:
….toml

This was added to resolve an issue between `flask_oidc` and `itsdangerous` which has since been resolved upstream. We don't actually depend on `itsdangerous` directly.


jcristau:
Ref: https://github.com/mozilla-releng/shipit/pull/1209#pullrequestreview-1194662456

jcristau (approved):

gbrownmozilla (approved):

gabrielBusta (approved):

Issue #1462, opened 2024-05-07
Stop hard-coding the Firefox Nightly version information in our configuration


gabrielBusta:
This issue pertains to the merge duty step outlined [here](https://docs.mozilla-releng.net/en/latest/how-to/releaseduty/merge-duty/merge_duty.html#bump-nightly-version-and-release-dates-in-shipit).

Currently, the Nightly version information is hard-coded into the shipit configuration file:

https://github.com/mozilla-releng/shipit/blob/01cc5f56a9121052377a0f0c7c9839492a50224b/api/src/shipit_api/common/config.py#L48-L55

This requires a manual source change every time there is a new Nightly version, followed by a shipit deployment. To reduce manual intervention from release engineering, we should enhance shipit to make it possible to update the current Nightly version without needing code modifications or application deployments.

This improvement will help automate this merge-day task and supports the objectives of [Issue #1418](https://github.com/mozilla-releng/shipit/issues/1418).

gabrielBusta:
```python
 # We could have used the in-tree version, but there can be race conditions, 
 # e.g. version bumped, but still no builds available
 ```
We could have an API to bump the version if needed from a shipitscript task in the Nightly graph. Kinda like bouncerscript.

PR #1465, opened 2024-05-09
Add a table to keep product channel versions and APIs to update them

gabrielBusta:
This is an API that lets us store the current versions of FIREFOX_NIGHTLY, instead of hard-coding it, so we can update this data using an API from a `shipitscript` task. This is part of the solution to #1462

### DIFF
--- a/api/src/shipit_api/admin/api.py
+++ b/api/src/shipit_api/admin/api.py
@@ -28,7 +28,7 @@ from shipit_api.admin.tasks import (
     rendered_hook_payload,
 )
 from shipit_api.common.config import HG_PREFIX, PROJECT_NAME, PULSE_ROUTE_REBUILD_PRODUCT_DETAILS, SCOPE_PREFIX
-from shipit_api.common.models import DisabledProduct, Phase, Release, Signoff, XPIRelease
+from shipit_api.common.models import DisabledProduct, Phase, Release, Signoff, Version, XPIRelease
 from shipit_api.public.api import get_disabled_products, list_releases
 
 logger = logging.getLogger(__name__)
@@ -375,3 +375,55 @@ def get_signoff_emails(phases):
             additional_shipit_emails.update({signoff.completed_by for signoff in _phase.signoffs if signoff.completed_by is not None})
 
     return list(additional_shipit_emails)
+
+
+def get_product_channel_version(product, channel):
+    version = Version.query.filter_by(product_name=product, product_channel=channel).first()
+    if version:
+        return version.current_version
+    else:
+        return {"error": f"No version found for {product} {channel}."}, 404
+
+
+def update_product_channel_version(product, channel, body):
+    required_permission = f"{SCOPE_PREFIX}/update_product_channel_version/{product}"
+    if not current_user.has_permissions(required_permission):
+        user_permissions = ", ".join(current_user.get_permissions())
+        abort(401, f"required permission: {required_permission}, user permissions: {user_permissions}")
+    version = Version.query.filter_by(product_name=product, product_channel=channel).first()
+    if version and version.current_version != body["version"]:
+        version.current_version = body["version"]
+        current_app.db.session.commit()
+        logger.info(f"Regenerating product details after updating {version.product_name} {version.product_channel} version to {version.current_version}")
+        _rebuild_product_details({})
+        notify_via_matrix(product, f"Updated {version.product_name} {version.product_channel} version to `{version.current_version}`.")
+        return {
+            "message": f"The version for {product} {channel} was updated successfully.",
+            "version": version.current_version,
+            "product": version.product_name,
+            "channel": version.product_channel,
+        }, 200
+    elif version and version.current_version == body["version"]:
+        return {"error": f"The {product} {channel} version is already {body['version']}!"}, 409
+    else:
+        return {"error": f"No version found for {product} {channel}."}, 404
+
+
+def create_product_channel_version(product, channel, body):
+    required_permission = f"{SCOPE_PREFIX}/create_product_channel_version/{product}"
+    if not current_user.has_permissions(required_permission):
+        user_permissions = ", ".join(current_user.get_permissions())
+        abort(401, f"required permission: {required_permission}, user permissions: {user_permissions}")
+    existing_version = Version.query.filter_by(product_name=product, product_channel=channel).first()
+    if existing_version:
+        return {"error": f"A {product} {channel} version already exists."}, 409
+    else:
+        new_version = Version(product_name=product, product_channel=channel, current_version=body["version"])
+        current_app.db.session.add(new_version)
+        current_app.db.session.commit()
+        return {
+            "message": f"A {product} {channel} version was created successfully.",
+            "version": new_version.current_version,
+            "product": new_version.product_name,
+            "channel": new_version.product_channel,
+        }, 201
--- a/api/src/shipit_api/admin/settings.py
+++ b/api/src/shipit_api/admin/settings.py
@@ -99,7 +99,8 @@ LDAP_GROUPS = {
 AUTH0_AUTH_SCOPES = assign_ldap_groups_to_scopes()
 
 # other scopes
-AUTH0_AUTH_SCOPES.update({"rebuild_product_details": LDAP_GROUPS["firefox-signoff"], "update_release_status": []})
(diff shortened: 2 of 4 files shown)


api/src/shipit_api/admin/api.py
```diff
@@ -375,3 +375,44 @@ def get_signoff_emails(phases):
...
+def get_product_version(product):
+    version = Version.query.filter_by(product_name=product).first()
+    if version:
+        return version.current_version
+    else:
+        return {"error": "Product version not found."}, 404
+
+
+def update_product_version(product, body):
+    # Only shipitscript should be able to call this API to update a product version.
```
gabrielBusta, line 389: This might be a little too simplistic. I think we might want to make it so we can have the scopes per product because I am pretty sure Thunderbird will want to use this instead of submitting a pull request every release day. Firefox Nightly is not a product in ShipIt (Firefox is.) Maybe we can prefix with a product? Also, all this is making me think we could use a Product model instead of hard-coding so much data in this service.

gabrielBusta: Ugh but I don't really want to go there. I want to keep to [KISS](https://en.wikipedia.org/wiki/KISS_principle). But, I definitely want to enable Thunderbird to use this API so we don't have to keep doing deployments on release day.

gabrielBusta:
To grant shipitscript scopes to call these APIs we'd need a patch in ci-configuration. Something like this:

```diff
diff --git a/clients.yml b/clients.yml
--- a/clients.yml
+++ b/clients.yml
@@ -862,6 +862,8 @@ project/releng/scriptworker/v2/shipit/de
     - project:releng:services/shipit_api/add_release/devedition
     - project:releng:services/shipit_api/add_release/fennec
     - project:releng:services/shipit_api/add_release/firefox
+    - project:releng:services/shipit_api/update_product_channel_version/firefox/nightly
+    - project:releng:services/shipit_api/create_product_channel_version/firefox/nightly
     - project:releng:services/shipit_api/dev/add_release/devedition
     - project:releng:services/shipit_api/dev/add_release/firefox
     - project:releng:services/shipit_api/dev/schedule_phase/devedition/ship_devedition
@@ -897,6 +899,8 @@ project/releng/scriptworker/v2/shipit/pr
     - project:releng:services/shipit_api/add_release/devedition
     - project:releng:services/shipit_api/add_release/fennec
     - project:releng:services/shipit_api/add_release/firefox
+    - project:releng:services/shipit_api/update_product_channel_version/firefox/nightly
+    - project:releng:services/shipit_api/create_product_channel_version/firefox/nightly
     - project:releng:services/shipit_api/dev/add_release/devedition
     - project:releng:services/shipit_api/dev/add_release/firefox
     - project:releng:services/shipit_api/dev/schedule_phase/devedition/ship_devedition
@@ -918,6 +922,8 @@ project/releng/scriptworker/v2/shipit/pr
     - project:releng:services/shipit_api/add_release/devedition
     - project:releng:services/shipit_api/add_release/fennec
     - project:releng:services/shipit_api/add_release/firefox
+    - project:releng:services/shipit_api/update_product_channel_version/firefox/nightly
+    - project:releng:services/shipit_api/create_product_channel_version/firefox/nightly
     - project:releng:services/shipit_api/production/add_release/devedition
     - project:releng:services/shipit_api/production/add_release/firefox
     - project:releng:services/shipit_api/production/schedule_phase/devedition/ship_devedition
```

api/src/shipit_api/admin/api.py
```diff
@@ -375,3 +375,52 @@ def get_signoff_emails(phases):
             additional_shipit_emails.update({signoff.completed_by for signoff in _phase.signoffs if signoff.completed_by is not None})
 
     return list(additional_shipit_emails)
+
+
+def get_product_channel_version(product, channel):
```
gabrielBusta, line 380: We should add this to the public OpenAPI Specification as well.

gabrielBusta:
This will need to be followed by https://github.com/mozilla-releng/shipit/pull/1489/commits/7fef0f61a8da5409740c4408afc6d407bbc4f8c7

gabrielBusta:
This needs to be followed by a commit https://github.com/mozilla-releng/shipit/pull/1489/commits/7fef0f61a8da5409740c4408afc6d407bbc4f8c7 to use the Firefox Nightly version stored in the database when rebuilding product details. The reason this has to land first is that I need to seed the initial Firefox Nightly version in the database. Otherwise product details will fail because the new table is empty.

api/src/shipit_api/admin/settings.py
```diff
@@ -99,7 +99,8 @@
 AUTH0_AUTH_SCOPES = assign_ldap_groups_to_scopes()
 
 # other scopes
-AUTH0_AUTH_SCOPES.update({"rebuild_product_details": LDAP_GROUPS["firefox-signoff"], "update_release_status": []})
+AUTH0_AUTH_SCOPES.update({"rebuild_product_details": LDAP_GROUPS["firefox-signoff"], "update_release_status": [], "create_product_channel_version/firefox": []})
```
gabrielBusta, line 102: Another follow-up: We will need to grant releng scopes to seed the initial Thunderbird version once the Thunderbird bits are implemented.

```suggestion
AUTH0_AUTH_SCOPES.update(
    {
        "rebuild_product_details": LDAP_GROUPS["firefox-signoff"],
        "update_release_status": [],
        "create_product_channel_version/firefox": [],
        "create_product_channel_version/thunderbird": [],
    }
)
```

gabrielBusta:
I ran a staging Nightly to try out this APIs in dev.

[Here's a run when the Nightly version has not changed.](https://firefox-ci-tc.services.mozilla.com/tasks/VBIiiBy1R760eHB7XTkbHg/runs/0/logs/public/logs/live_backing.log#L9)

[Here's a run with a simulated Nightly version bump.](https://firefox-ci-tc.services.mozilla.com/tasks/DjER4oGLR1u9f1xkN6KHbw/runs/0/logs/public/logs/live_backing.log#L9)

api/src/shipit_api/admin/api.py
```diff
@@ -375,3 +375,55 @@ def get_signoff_emails(phases):
...
+    version = Version.query.filter_by(product_name=product, product_channel=channel).first()
+    if version:
+        return version.current_version
```
hneiva, line 383: nit: All other functions here seem to return a tuple with (data, exitcode) should we follow the same pattern here?

api/src/shipit_api/admin/api.py
```diff
@@ -375,3 +375,55 @@ def get_signoff_emails(phases):
...
+    else:
+        return {"error": f"No version found for {product} {channel}."}, 404
+
+
+def update_product_channel_version(product, channel, body):
+    required_permission = f"{SCOPE_PREFIX}/update_product_channel_version/{product}"
+    if not current_user.has_permissions(required_permission):
+        user_permissions = ", ".join(current_user.get_permissions())
+        abort(401, f"required permission: {required_permission}, user permissions: {user_permissions}")
```
hneiva, line 392: nit: What's the difference between aborting and returning an error?
Also, seems like (data, exitcode) should be the pattern used?

api/src/shipit_api/admin/api.py
```diff
@@ -375,3 +375,55 @@ def get_signoff_emails(phases):
...
+        return {"error": f"The {product} {channel} version is already {body['version']}!"}, 409
+    else:
+        return {"error": f"No version found for {product} {channel}."}, 404
+
+
+def create_product_channel_version(product, channel, body):
+    required_permission = f"{SCOPE_PREFIX}/create_product_channel_version/{product}"
+    if not current_user.has_permissions(required_permission):
+        user_permissions = ", ".join(current_user.get_permissions())
+        abort(401, f"required permission: {required_permission}, user permissions: {user_permissions}")
```
hneiva, line 416: same as above

api/src/shipit_api/admin/api.py
```diff
@@ -375,3 +375,55 @@ def get_signoff_emails(phases):
...
+    existing_version = Version.query.filter_by(product_name=product, product_channel=channel).first()
+    if existing_version:
```
hneiva, line 418: nit: could improve readability here:
```
if Version.query.filter_by(product_name=product, product_channel=channel).first():
    return {"error": f"A {product} {channel} version already exists."}, 409
# code here without else
```

hneiva (approved):
LGTM - just a few nits/questions, but nothing blocking this from merging

gabrielBusta:
> LGTM - just a few nits/questions, but nothing blocking this from merging

Thanks, I will address on a follow-up

PR #1475, opened 2024-05-16
Setup `shipit_mozillavpn` LDAP access to VPN products

ahal:


api/src/shipit_api/admin/settings.py
```diff
@@ -80,7 +80,7 @@
     "thunderbird-signoff": ["shipit_thunderbird"],
     "firefox-android-signoff": ["shipit_mobile"],
     "app-services-signoff": ["shipit_app_services"],
-    "vpn-signoff": ADMIN_LDAP_GROUP,
+    "vpn-signoff": ["shipit_mozillavpn"],
```
ahal, line 83: I thought about keeping `ADMIN_LDAP_GROUP` here, but I worry we'll abuse the power. Besides in a true emergency, we'd probably shutdown updates in Balrog first anyway.

gabrielBusta (approved):

api/src/shipit_api/admin/settings.py
gabrielBusta: True that

PR #1488, opened 2024-05-30
Add Copy Access Token item to the user menu

### LINKED ISSUE #1487: Add "Copy Access Token" item to the user menu
Balrog has this nice feature where there's a handy menu item to copy the access token for use with CLI tools and scripts. Let's add it to shipit too.

gabrielBusta:
Resolves #1487


JohanLorenzo (approved):
I'm no react expert but this PR makes sense to me. Thanks!

PR #1489, opened 2024-05-31
Use the firefox nightly version in db

gabrielBusta:
This is a follow up to #1465 

Once that one lands, I need to seed the initial Firefox Nightly version in the DB, and from then on we should use that value when we rebuild product details.


gabrielBusta:
This is ready for review. But I need to deploy #1465 and seed the initial version before landing it.

gbrownmozilla (approved):

gabrielBusta:
#1462 1462

gabrielBusta:
#1462

PR #1493, opened 2024-06-04
Python dependency updates

bhearsum:


bhearsum:
Looks like dependabot did a bunch already this week; foregoing this

PR #1498, opened 2024-06-10
Assume the next Firefox release will be the day after the last merge date unless the data says otherwise

gabrielBusta:
i.e. fallback to the day after the last merge date if there has not been a Firefox release since.


gbrownmozilla (approved):
Thanks!

PR #1523, opened 2024-07-10
Bug 1884740 - handle "auto" partials for firefox-android

jcristau:


jcristau:
I still want to add tests for this, but am not sure it's worth holding up this fix.

api/src/shipit_api/admin/api.py
```diff
@@ -70,7 +70,7 @@ def add_release(body):
     session = current_app.db.session
     partial_updates = body.get("partial_updates")
     if partial_updates == "auto":
-        if product not in [Product.FIREFOX.value, Product.DEVEDITION.value] or branch not in [
+        if product not in (Product.FIREFOX.value, Product.DEVEDITION.value, Product.FIREFOX_ANDROID.value) or branch not in [
```
bhearsum, line 73: Maybe a silly question, but why is `partial_updates` set at all when this is called for Android?

jcristau: Not silly at all.

The client side at https://github.com/mozilla-releng/scriptworker-scripts/blob/608e8228bf2c001f954a0f4cc01464a3d35e50b0/shipitscript/src/shipitscript/shipitapi.py#L122-L124 is arguably what should be fixed instead of this...

gabrielBusta (approved):

PR #1525, opened 2024-07-12
ci: for pushes, only run taskcluster on actively used branches

bhearsum:
Anything else should route through a PR. If we need other long lived branches in the future we can add them, and also add them to https://github.com/mozilla-releng/fxci-config/blob/main/projects.yml.


jcristau (approved):
Thanks!

PR #1531, opened 2024-07-22
Accept the empty string as a falsy value for build_number in list_releases

### LINKED ISSUE #1530: Permanent 400/white screen crash blocking the creation of Thunderbird releases
### Summary

**API Endpoint**: `GET /releases`

**Error**: `"Wrong type, expected 'integer' for query parameter 'build_number'"`

Note: This might be a regression caused by [updating axios](https://github.com/mozilla-releng/shipit/commit/91e98dfff934c3f1020e156f2690220a6a77389f), which may have changed how it handles null/undefined parameters.

https://github.com/mozilla-releng/shipit/blob/91e98dfff934c3f1020e156f2690220a6a77389f/frontend/src/components/api.js#L77

https://github.com/mozilla-releng/shipit/blob/91e98dfff934c3f1020e156f2690220a6a77389f/frontend/src/components/api.js#L13-L24

### Description

When making a request to the `GET /releases` endpoint, the server returns a `400 Bad Request` error if the `build_number` query parameter is provided as an empty string.

### Steps to Reproduce

1. Send a request to the `/releases` endpoint with the `build_number` query parameter as an empty string:
   ```
   GET /releases?product=thunderbird&branch=releases%2Fcomm-beta&version=&build_number=&status=shipped
   ```
2. Get a `400 Bad Request` response:
   ```json
   {
     "detail": "Wrong type, expected 'integer' for query parameter 'build_number'",
     "status": 400,
     "title": "Bad Request",
     "type": "about:blank"
   }
   ```

### Expected Behavior

The server should handle the empty string for the `build_number` parameter gracefully by interpreting it as if the parameter is not provided.

gabrielBusta:
Fixes #1530

### DIFF
--- a/api/src/shipit_api/public/api.py
+++ b/api/src/shipit_api/public/api.py
@@ -30,7 +30,9 @@ def good_version(release):
         return False
 
 
-def list_releases(product=None, branch=None, version=None, build_number=None, status=["scheduled"]):
+def list_releases(product=None, branch=None, version=None, build_number=None, status=None):
+    if not status:
+        status = ["scheduled"]
     session = current_app.db.session
     releases = session.query(Release)
     if product:
@@ -40,7 +42,7 @@ def list_releases(product=None, branch=None, version=None, build_number=None, st
     if version:
         releases = releases.filter(Release.version == version)
         if build_number:
-            releases = releases.filter(Release.build_number == build_number)
+            releases = releases.filter(Release.build_number == int(build_number))
     elif build_number:
         raise BadRequest(description="Filtering by build_number without version is not supported.")
     releases = releases.filter(Release.status.in_(status))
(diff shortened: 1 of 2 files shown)


gbrownmozilla (approved):
I'm not sure I understand the issue, but this seems worth trying.

api/src/shipit_api/public/api.py
```diff
@@ -40,7 +42,7 @@ def list_releases(product=None, branch=None, version=None, build_number=None, st
     if version:
         releases = releases.filter(Release.version == version)
         if build_number:
-            releases = releases.filter(Release.build_number == build_number)
+            releases = releases.filter(Release.build_number == int(build_number))
```
jcristau, line 45: If build_number is not an int this will raise ValueError and presumably turn into a 500, which we don't want.

jcristau (changes requested):
This feels like it should be fixed in the frontend, to not send a null value...

PR #1541, opened 2024-08-19
remove gbrownmozilla from CODEOWNERS

gbrownmozilla:


hneiva (approved):
We should probably add a second person as backup later.

PR #1559, opened 2024-09-25
api: set CURRENT_ESR to 128

jcristau:


jcristau:
This should be deployed to prod on October 1st.

bhearsum (approved):

PR #1585, opened 2024-12-09
Bump local dev db to postgresql 15 to match prod

jcristau:


bhearsum (approved):

jcristau:
Added to the commit message: "Note this will require something like `docker volume rm shipit_dbstore` in local dev environments, since the db container will refuse to start against an old datadir."

PR #1593, opened 2024-12-16
product-details: sanitize sanity_check_firefox_primary_builds

jcristau:
There's no use calling sanity_check_firefox_primary_builds with a version key that it immediately discards.


bhearsum (approved):
:lol:

PR #1595, opened 2024-12-17
Add test coverage for product-details rebuild

jcristau:
Prompted by recent breakage; this is based on #1594.

### DIFF
--- a/api/tests/test_product_details.py
+++ b/api/tests/test_product_details.py
@@ -3,16 +3,33 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import json
 import pathlib
 import re
+import subprocess
+from unittest import mock
 
 import aiohttp
 import pytest
 from aioresponses import aioresponses
+from sqlalchemy import engine, event
 
 import shipit_api.admin.product_details
-from shipit_api.admin.product_details import fetch_l10n_data
-from shipit_api.common.models import Release
+from shipit_api.admin.product_details import fetch_l10n_data, rebuild
+from shipit_api.common.models import Release, Version
+
+
+# product_details uses a postgresql-specific "split_part" sql function
+@event.listens_for(engine.Engine, "connect")
+def setup_split_part(dbapi_connection, conn_rec):
+    def split_part(string, delimiter, position):
+        return string.split(delimiter, position)[position - 1]
+
+    dbapi_connection.create_function(
+        "split_part",
+        3,
+        split_part,
+    )
 
 
 def create_html(folder, items):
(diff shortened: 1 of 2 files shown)


api/tests/test_product_details.py
```diff
@@ -72,3 +87,74 @@ async def test_fetch_l10n_data():
         m.get(url, status=200, payload=dict(a="a"))
         (_, changesets) = await fetch_l10n_data(session, release, raise_on_failure=True, use_cache=False)
         assert changesets == {"a": "a"}
+
+
+def mock_setup_working_copy(branch, url, secrets):
+    subprocess.check_call(["git", "clone", "-n", url, str(shipit_api.common.config.PRODUCT_DETAILS_DIR)])
```
jcristau, line 93: This is not great, the test should be self-contained, but it was easiest for now.  There's a few other network requests done as part of the rebuild that should be mocked as well...

bhearsum: Not ideal, but anything is better than nothing at this point...

api/tests/test_product_details.py
```diff
@@ -72,3 +89,79 @@ async def test_fetch_l10n_data():
...
+    assert versions["FIREFOX_NIGHTLY"] == "135.0a1"
+    assert versions["FIREFOX_DEVEDITION"] == "134.0b9"
+    assert versions["LATEST_FIREFOX_DEVEL_VERSION"] == "134.0b8"
+    assert versions["LATEST_FIREFOX_VERSION"] == "133.0"
+
+    with (parent / "firefox_primary_builds.json").open() as f:
+        primary_builds = json.load(f)
+    assert set(primary_builds["ach"].keys()) == {"133.0", "134.0b8", "134.0b9", "135.0a1"}
+
+    assert not list(parent.glob("l10n/Devedition-*"))
```
bhearsum, line 166: ...and this covers #1596, I assume?

api/tests/test_product_details.py
```diff
@@ -72,3 +89,79 @@ async def test_fetch_l10n_data():
...
+        revision="615791f9752c70ef1757abf68544c8275f219ce3",
+        build_number=1,
+        release_eta=None,
+        status="shipped",
+        partial_updates=None,
+    )
+    beta = Release(
+        product="firefox",
+        branch="releases/mozilla-beta",
+        version="134.0b8",
```
bhearsum, line 124: If I understand correctly, this + having a newer DevEdition version should cover the issue that caused the original backout?

bhearsum (approved):
No doubt there's more tests that could be added here, but I don't see any reason to block this on adding them.

api/tests/test_product_details.py
gabrielBusta: Maybe there could be a static data directory under `tests`? Then you could store a snapshot of product details data in there and init a git repo in there at test run-time. I think at that point you wouldn't need to mock the function (can probably use the local repo path as the `url`)

The raw data is about 32M:

```
git clone https://github.com/mozilla-releng/product-details.git
Cloning into 'product-details'...
... done.
du -sh product-details
 44M	product-details
rm -rf product-details/.git
du -sh product-details
 32M	product-details
 ```
 
 The network calls to HG for the l10n data are more likely to fail/hit rate limits. Those would be good to mock.

api/tests/test_product_details.py
gabrielBusta: Nice.

gabrielBusta (approved):

api/tests/test_product_details.py
jcristau: I think I'll land this before I run out of steam, and try and improve later.  We can probably get away with a lot less than 32M (maybe even git init an empty repo) and still get meaningful testing, but I'd rather not block on that given you two already reviewed this.

PR #1622, opened 2025-02-12
feat: Bug 1933778 - Add Firefox iOS

hneiva:
Missing:
- [x] Fix release promotion action in decision task


frontend/src/configs/production.js
```diff
@@ -101,6 +101,26 @@ module.exports = {
...
+        prettyName: 'Firefox iOS',
+        appName: 'firefox-ios',
+        branches: [
+          {
+            branch: '',
+          },
+        ],
+        repositories: [
+          {
+            prettyName: 'Firefox iOS Staging',
```
Eijebong, line 115: Should just be `Firefox iOS`

api/src/shipit_api/admin/github.py
```diff
@@ -226,6 +226,9 @@ def get_package_json_directory(owner, repo, revision, directory):
 
 
 def get_version_txt(owner, repo, revision):
+    if "firefox-ios" in repo:
+        # TODO: Figure out what to do about version
+        return "900.0"
```
Eijebong, line 231: Could we parse it from `firefox-ios/Client/Info.plist` instead? https://github.com/mozilla-mobile/firefox-ios/commit/3c68659ab379fbf76cb57dcc5b280079d3b73303 the version is specified in a bunch of places but AFAIK you cannot set PList fields based on other files.

Should be fairly easy too because for some reason https://docs.python.org/3/library/plistlib.html is stdlib

Eijebong: A version.txt file has been added.

api/src/shipit_api/admin/settings.py
```diff
@@ -79,6 +79,7 @@
     "firefox-signoff": ["shipit_firefox"],
     "thunderbird-signoff": ["shipit_thunderbird"],
     "firefox-android-signoff": ["shipit_mobile"],
+    "firefox-ios-signoff": ["shipit_firefox_ios"],
```
hneiva, line 82: should this be `shipit_mobile` as well?

Eijebong (approved):

PR #1626, opened 2025-02-14
Create action decisions and task groups with titles that say what release phase they correspond to

### LINKED ISSUE #1625: Create action decisions and task groups with titles that say what release phase they correspond to
It's difficult to keep Taskcluster tabs straight when monitoring/debugging releases because they all say "Action: Release Promotion"

ShipIt might be able to pass in more information via. the generated phase hooks.

gabrielBusta:
Fixes #1625 

The action decisions looks like this:
Screenshot 2025-02-14 at 15 33 16

The resulting task groups looks like this:
Screenshot 2025-02-14 at 15 33 45

### DIFF
--- a/api/src/shipit_api/admin/tasks.py
+++ b/api/src/shipit_api/admin/tasks.py
@@ -157,10 +157,12 @@ def extract_our_flavors(avail_flavors, product, version, partial_updates, produc
     return [flavor for flavor in SUPPORTED_FLAVORS[product_key] if flavor["name"] in avail_flavors_set]
 
 
-def generate_action_hook(task_group_id, action_name, actions, parameters, input_):
+def generate_action_hook(task_group_id, action_name, actions, parameters, input_, action_title=None):
     target_action = find_action(action_name, actions)
     context = copy.deepcopy({"parameters": parameters})
     context.update({"taskGroupId": task_group_id, "taskId": None, "task": None, "input": input_})
+    if action_title:
+        target_action["hookPayload"]["decision"]["action"]["title"] = action_title
     return dict(hook_group_id=target_action["hookGroupId"], hook_id=target_action["hookId"], hook_payload=target_action["hookPayload"], context=context)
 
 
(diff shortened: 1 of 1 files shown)


bhearsum (approved):
This seems like it ought to be fine. It's _slightly_ weird to be overriding the existing `title`, but we're getting a lot of benefit from doing so....so meh.

gabrielBusta:
> This seems like it ought to be fine. It's _slightly_ weird to be overriding the existing `title`, but we're getting a lot of benefit from doing so....so meh.

It is... kind of surprising. This hack is overriding [the value for `"title"` defined using `taskgraph` actions.](https://github.com/taskcluster/taskgraph/blob/c49b05d003f989e2fbba229308c16cd06b7127bf/src/taskgraph/actions/registry.py#L84-L86) Those values are [written in the decision task to the `actions.json` file.](https://github.com/taskcluster/taskgraph/blob/c49b05d003f989e2fbba229308c16cd06b7127bf/src/taskgraph/decision.py#L118-L121) `shipit` [renders the hook payload using that data](https://github.com/mozilla-releng/shipit/blob/bc706a2cc2cee44b4a60a642333af87a7daaf3ea/api/src/shipit_api/admin/tasks.py#L226-L227) and saves it in the database as part of the `Phase`. That's the data that gets passed into the `in-tree-action-{level}-release-promotion/{hash}` hook on `TRIGGER_HOOK`. [In the hook's "Task Template"](https://firefox-ci-tc.services.mozilla.com/hooks/project-gecko/in-tree-action-1-release-promotion%2F1ae5097b9b) I see the mapping on `tasks_for == "action"` from `payload.decision.action.title` to the task's `metadata.name` field. [This seems to be a display value for users](https://github.com/taskcluster/taskcluster/blob/5e78a283c77938b2c92f7c82c866c166017abed4/services/queue/schemas/v1/task-metadata.yml#L9-L11), so this _should_ be OK... that being said... I'll do a little bit more testing (create some releases for `.xpi`s, `app-services`, etc.)

gabrielBusta:
I did some testing and things look OK. Worse case scenario, I think, we get some weird group names.

PR #1628, opened 2025-02-21
Create an issue template to point people to Bugzilla

JohanLorenzo:
If we turn off the Issues feature on the repo, then links of all previously created issues become 404. We still want to keep the history around.


gabrielBusta:
better than nothing

gabrielBusta (approved):

PR #1637, opened 2025-03-03
Add mock to the mocked client secrets file name and clean up old/unnecessary testing data

gabrielBusta:


gabrielBusta:
> So we didn't even real secrets to run this tests? 🤣

Nope 🤣

PR #1643, opened 2025-03-28
Bug 1949703 - Automatically refresh the status of release phases in pending releases

gabrielBusta:
https://bugzilla.mozilla.org/show_bug.cgi?id=1949703


frontend/src/components/PhaseProgress/index.jsx
```diff
@@ -130,7 +140,7 @@ export default function PhaseProgress({ release, readOnly, xpi }) {
                 : { classes: { completed: classes.completed } }
             }>
             <Link href={`${taskGroupUrlPrefix}/${phase.actionTaskId}`}>
-              {prettyName} task
+              {prettyName} task group
```
gabrielBusta, line 143: Sometimes this is a 404. I wonder if we could wait for it to actually be there before making this link clickable

PR #1649, opened 2025-04-07
Bug 1951760 - Support an LDAP group per System Addon

ahal:
This implements Releng RFC 0050:
https://github.com/mozilla-releng/releng-rfcs/blob/66e0a531bd7c3103ea2d985ac449830caad0c181/rfcs/0050-system-addons-release-process.md

### DIFF
--- a/api/src/shipit_api/admin/settings.py
+++ b/api/src/shipit_api/admin/settings.py
@@ -4,12 +4,13 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import base64
+from itertools import chain
 
 from decouple import config
 
 from backend_common.auth import create_auth0_secrets_file
 from shipit_api.admin.auth0 import assign_ldap_groups_to_scopes
-from shipit_api.common.config import SCOPE_PREFIX, XPI_LAX_SIGN_OFF
+from shipit_api.common.config import SCOPE_PREFIX, SYSTEM_ADDONS, XPI_LAX_SIGN_OFF
 
 # TODO: 1) rename "development" to "local" 2) remove "staging" when fully migrated
 supported_channels = ["dev", "development", "staging", "production"]
@@ -68,8 +69,6 @@ ADMIN_LDAP_GROUP = ["releng"]
 XPI_PRIVILEGED_BUILD_LDAP_GROUP = ["xpi_privileged_build"]
 
 XPI_PRIVILEGED_ADMIN_LDAP_GROUP = ["xpi_privileged_admin"]
-XPI_SYSTEM_BUILD_LDAP_GROUP = ["xpi_system_build"]
-XPI_SYSTEM_ADMIN_LDAP_GROUP = ["xpi_system_admin"]
 XPI_MOZILLAONLINE_PRIVILEGED_LDAP_GROUP = ["xpi_mozillaonline_build"]
 XPI_MOZILLAONLINE_PRIVILEGED_ADMIN_LDAP_GROUP = ["xpi_mozillaonline_admin"]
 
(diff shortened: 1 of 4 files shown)


ahal:
This seems to work from my local testing, but it doesn't implement the entire RFC. Apart from fixing the test, the remaining work is:

1. Remove the 'promote' phase. Afaict, phases are parsed from the action task definitions, so this will involve a change to `xpi-manifest`. We'll also want to move the signing tasks over to the "ship" phase as part of this.
2. Create the `shipit_xpi_system_webcompat` LDAP group.

api/src/shipit_api/admin/settings.py
```diff
@@ -89,12 +88,12 @@
...
     "xpi_privileged_signoff": XPI_PRIVILEGED_ADMIN_LDAP_GROUP + ADMIN_LDAP_GROUP,
     "xpi_privileged_admin_signoff": XPI_PRIVILEGED_ADMIN_LDAP_GROUP,
-    "xpi_system_build": XPI_SYSTEM_BUILD_LDAP_GROUP + ADMIN_LDAP_GROUP,
-    "xpi_system_signoff": XPI_SYSTEM_ADMIN_LDAP_GROUP + ADMIN_LDAP_GROUP,
     "xpi_mozillaonline-privileged_signoff": XPI_MOZILLAONLINE_PRIVILEGED_LDAP_GROUP + ADMIN_LDAP_GROUP,
     "xpi_mozillaonline-privileged_admin_signoff": XPI_MOZILLAONLINE_PRIVILEGED_ADMIN_LDAP_GROUP + ADMIN_LDAP_GROUP,
     "xpi_normandy-privileged_signoff": ADMIN_LDAP_GROUP,
 }
+for addon in XPI_SYSTEM_ADDONS:
+    LDAP_GROUPS[f"xpi_system_{addon}"] = [f"shipit_xpi_system_{addon}"] + ADMIN_LDAP_GROUP
```
ahal, line 96: Hm, if we're going to move system addons out of the `mozilla-extensions` org and into Gecko, maybe we at least shouldn't have `xpi` in the LDAP group name (as that will be more of a pain to change later).

ahal:
> Remove the 'promote' phase. Afaict, phases are parsed from the action task definitions, so this will involve a change to xpi-manifest. We'll also want to move the signing tasks over to the "ship" phase as part of this.

This is actually a bit trickier than I anticipated.. The phases are determined by parsing the actions schema from the `xpi-manifest` repo. The phases are defined in `config.yml` there, and processed at import time in `actions/release_promotion.py` (i.e in the action decorator).

Therefore there's no real ability to have a different set of phases based on the addon type being shipped, we'd need to have a whole new relpro action for system addons. This is definitely do-able, but it would require significant changes to both `xpi-manifest` and `shipit`.

Since the main goal here is to focus on shipping system addons out of m-c, I'm planning to punt any changes to the # of phases in the "old" way of shipping system addons.

gabrielBusta:
Weird. I'd have guessed [api/products.yml](https://github.com/mozilla-releng/shipit/blob/ac406fbc91bfe03654aa54e55db3f17d5ac330a4/api/products.yml)

gabrielBusta:
Yeah this would require some refactoring. Maybe the the config.yml in xpi manifest could be more like product.yml

Like:
 ```
release-promotion:
    default:
        build:
            target-tasks-method: build_xpi
            rebuild-kinds:
                - docker-image
                - build
                - test
                - addons-linter
                - dep-signing
        promote:
            target-tasks-method: promote_xpi
        ship:
            target-tasks-method: ship_xpi
    system:
        build:
            target-tasks-method: build_xpi
            rebuild-kinds:
                - docker-image
                - build
                - test
                - addons-linter
                - dep-signing
        ship:
            target-tasks-method: ship_xpi
 ```

api/src/shipit_api/common/config.py
```diff
@@ -453,6 +453,7 @@ def _get_phases_definitions(phases):
 
 SUPPORTED_FLAVORS = _get_supported_flavors()
 
+SYSTEM_ADDONS = ("webcompat",)
```
gabrielBusta, line 456: What about newtab?

ahal: Was planning to do that in a follow-up because at the time I wrote this patch, it didn't exist. I guess I could add it, but it might be nice to have a patch to point to on how to add a new one of these things.

gabrielBusta (approved):

PR #1650, opened 2025-04-16
Bug 1961228: Use webpack instead of Neutrino to build the shipit UI

gabrielbusta:
Using Neutrino to build the shipit UI blocks updating babel/loaders/react. The project is been abandoned since 2020. These changes deprecate Neutrino and set up the shipit UI as a standard webpack project. The dependencies pulled in by Neutrino under the hood are pinned to the version the were at before the migration, to reduce the changes introduced in this PR. Once it lands and builds, it will be possible to start updating some of shipit's outdated dependencies.


gabrielBusta:
The node on the application's image is also out-dated, so that will likely need to be updated to be able to run newer versions of the packages.

ahal (approved):
So I don't really have any experience with either neutrino or webpack.. but I'm guessing no one on the team does, so I might as well rubber stamp it.

Let's get this deployed to dev and we can take it for a spin!

gabrielBusta:
@ahal its deploying to dev

ahal (changes requested):
Requesting changes until we investigate / sort out those failures.

gabrielbusta:
The first one looks like a back-end error. I will inspect them.

ahal (approved):
Thanks for doing that.. I agree we should try and get this landed before it starts to bitrot. If things are looking good in dev, I guess we should just give it a shot?

Btw this is an absolute rubber stamp.

PR #1653, opened 2025-04-21
Bug 1961246 - Add 'newtab' system addon to shipit

ahal:
Unfortunately the per system addon LDAP groups require us to explicitly list system addons in shipit. It might be worth a follow-up refactoring to try and avoid this requirement. However there is also going to be a larger refactor around system addons and shipping them out of mozilla-central. So I think it's ok to just require a shipit deploy in the meantime.


gabrielBusta (approved):

gabrielBusta:
Yeah true. There's also not that many of them.
Maybe `shipit` can use the addon type from the xpi manifest to filter for `system` (or something like that)

PR #1662, opened 2025-05-08
refactor: merge several github endpoints into /github/file

ahal:
There are at least three endpoints that fetch specific files from Github. Simplify these by providing a generic `/github/file` endpoint that can be used instead.


bhearsum (changes requested):
Thanks for taking the time to refactor this!

It looks like the CI failures are just linting (I think around the poetry lock file); I'm marking as request changes for that, but the patch itself looks fine.

bhearsum (approved):

PR #1664, opened 2025-05-13
Fixes and optimizations to `github/file` endpoint

ahal:


ahal:
I missed that we also fetch this file in my last PR as I was only looking at the public endpoints.

bhearsum (approved):

PR #1668, opened 2025-05-15
Set `asyncio_default_fixture_loop_scope` to silence a bunch of warnings

Eijebong:
Not setting this is deprecated (see
https://github.com/pytest-dev/pytest-asyncio/issues/924). We don't even use async fixtures AFAICT so it doesn't impact us but it still emits a lot of useless warnings

### DIFF
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -108,9 +108,11 @@ requests = "*"
 [tool.poetry.group.tox.dependencies]
 tox = "*"
 
-
 [tool.poetry.group.dev.dependencies]
 flask-shell-ipython = "^0.5.1"
+
+[tool.pytest.ini_options]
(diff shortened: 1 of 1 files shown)


api/pyproject.toml
```diff
@@ -108,10 +108,12 @@ requests = "*"
 [tool.poetry.group.tox.dependencies]
 tox = "*"
 
-
 [tool.poetry.group.dev.dependencies]
 flask-shell-ipython = "^0.5.1"
 
+[tool.pytest.ini_options]
+asyncio_default_fixture_loop_scope="function"
```
jcristau, line 115: nit:
```suggestion
asyncio_default_fixture_loop_scope = "function"
```

jcristau (approved):

PR #1676, opened 2025-05-27
Add firefox ios disableable branches

Eijebong:
This was a lot more work than I expected because shipit doesn't really support having dynamic branch names for release channels.

My first attempt at this was to make the `branch` field in the configuration a regex because I couldn't find any other use for that field. Turns out it's used in the frontend for the release creation form where it will grab branches from the config to give them to the user iif the `branch` field isn't empty. Which means that I broke everything for firefox-ios because of course it would give me the branch `Try` that would then map to the regex and boom...

Instead of having a full rewrite of the config to differentiate release channels from branch names, I chose to consider that if the `empty string` branch is marked as disabled then any branch would get considered disabled. It's a bit ugly if you look at the API calls to disable/enable the releases but at least it works without breaking everything else.

### DIFF
--- a/api/src/shipit_api/admin/api.py
+++ b/api/src/shipit_api/admin/api.py
@@ -340,8 +340,13 @@ def disable_product(body):
     dp = DisabledProduct(product=product, branch=branch)
     session.add(dp)
     session.commit()
-    logger.info("Disabled %s on branch %s", product, branch)
-    notify_via_matrix(product, f"Automatic releases disabled for {product} on {branch}")
+
+    if branch == "":
+        logger.info("Disabled %s on all branches", product)
+        notify_via_matrix(product, f"Automatic releases disabled for {product} on all branches")
+    else:
+        logger.info("Disabled %s on branch %s", product, branch)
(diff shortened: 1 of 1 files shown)


Eijebong:
First run with releases disabled for firefox-ios, second one after re-enabling. https://firefox-ci-tc.services.mozilla.com/tasks/S6rlLgf9QxK3Eo9_N0GxFQ

bhearsum (commented):
I'm not super happy with the empty string, but without asking you to dive deeper than you probably ought to for this, I don't have much in the way of suggestions.

I am curious what this will look like on the frontend though - does it render nicely with an empty string, or do we end with "Firefox iOS: [checkmark or X]"?

Eijebong:
It only ever displays the pretty name on the UI

image
image

bhearsum:
> It only ever displays the pretty name on the UI
> 
> image image

What does this look like for production?

Eijebong:
> What does this look like for production?

Should be the same with `Beta` instead of `Try`

jcristau (approved):
Not completely satisfying but I also don't have a better suggestion.

PR #1700, opened 2025-06-30
Add a way to start the worker by calling init.sh

Eijebong:
Right now the deployment is hardcoded to running `poetry run flask worker` which is a bit of a pain if we want to switch to uv. By changing the entry point to use `init.sh` we can avoid breaking everything by merging this first to dev and prod, then changing the entrypoint in the deployment to use it. Then we're free to switch it to be using uv instead.


Eijebong:
Deployment PR: https://github.com/mozilla-services/cloudops-infra/pull/6487

PR #1714, opened 2025-07-17
Update webpack and related dependencies as well as node

Eijebong:
"I'm just going to update webpack today, it's going to be one dependency and it's going to be fine, we'll do things slowly" - Me 3h ago

Webpack required me to update webpack-cli/webpack-dev-server which makes sense. There was some babel error which forced me to update it which forced me to update node. Which started failing because of SSL related things (`ERR_OSSL_EVP_UNSUPPORTED`) in loaders like babel, mini-css-extract-plugin, file-loader, url-loader as well as mozilla-protocol. So I had to update these as well...

In the end this ended up being way bigger than I wanted but changes are very confined.

A quick note, I didn't update node directly to the latest version to try and keep things slow moving and avoid breaking everything at once... I'll update it again in a followup once we know that this commit doesn't break anything.

First, they seem to have changed how the mode/env is passed into the config function. I couldn't find anything about that change... But I confirmed that `mode` was always what was passed on the command line, so I just went with it, removed --env and used --mode instead.

The devtool name changed format, it's just a reordering of components, nothing else.

Now for the painful part. webpack stopped polyfilling in for node by default, which means that we have to instruct it to do that in `resolve.fallback` except when we don't and have to use a plugin instead... `assert` is required by at least `taskcluster-lib-urls`, and `process` is used to pass environment variables around for configuration. The former is a resolver fallback, the latter a plugin. Don't ask me why that's different unless you want me to stare at you with a blank look and shrug.

Babel also stopped working with webpack 5, thankfully https://github.com/webpack/webpack/issues/11467#issuecomment-691873586 solved that...

Finally file/url-loader are deprecated and can be replaced easily


Eijebong:
This is now deployed on staging for easier testing

bhearsum:
>  i-d-be-pulling-my-hair-off-so-bad-right-now-if-i-had-any-left 

:upside_down_face:

ahal (approved):
I'm so sorry.. but thanks for doing this!

jcristau (approved):

PR #1733, opened 2025-07-25
docker: merge Dockerfile and Dockerfile.public

jcristau:
The only difference between them is the command.


Eijebong (approved):
https://bugzilla.mozilla.org/show_bug.cgi?id=1949684

jcristau:
Thanks for the pointer.  Looks like that was suggesting going one step further and having a single image, not just a single Dockerfile, so I'll leave it open.

PR #1740, opened 2025-07-28
 Remove the dependency on mozilla-frontend-infra/components 

Eijebong:
It's been archived since 2021, it's stuck on material ui 4, and if we ever want to update react/material ui in shipit, we have to get rid of this.

We were only using 2 components from it, loading spinners which were easy enough to replace and the `MuiErrorPanel` which I could just replace with a basic `Alert`.

The spinner change shouldn't change anything display wise. The switch to `Alert` makes the following change:

Before
image

After
image


jcristau (approved):
:broom:

PR #1745, opened 2025-07-28
frontend: change default port for local dev from 8010 to 8014

jcristau:
Avoid conflicting with a port used by balrog localdev.


jcristau:
The auth0 config expects localhost:8010 as redirect uri so this doesn't work; closing.

PR #1747, opened 2025-07-30
Update react to 19 and MUI to 7

Eijebong:
This also required node 20 because of react-router


bhearsum (commented):
> That matches something that I've seen in taskcluster after updates. It might be webpack related but it's definitely something new somehow, refreshing the page clears it out.

I don't think we should take this if it introduces a new intermittent error in production; can we hold on this until this is rooted out?

Eijebong:
> > That matches something that I've seen in taskcluster after updates. It might be webpack related but it's definitely something new somehow, refreshing the page clears it out.
> 
> I don't think we should take this if it introduces a new intermittent error in production; can we hold on this until this is rooted out?

https://github.com/mozilla-releng/shipit/pull/1747/files#diff-eb6063fd4b754545b1a874dc0b922bb0fdf014ddc7a9f0eba7e058c8f157be05R231

99% sure that this fixes it by forcing cache busting when the hash of the file changes. At least I have not seen it since I turned that on

bhearsum:
> > > That matches something that I've seen in taskcluster after updates. It might be webpack related but it's definitely something new somehow, refreshing the page clears it out.
> > 
> > 
> > I don't think we should take this if it introduces a new intermittent error in production; can we hold on this until this is rooted out?
> 
> https://github.com/mozilla-releng/shipit/pull/1747/files#diff-eb6063fd4b754545b1a874dc0b922bb0fdf014ddc7a9f0eba7e058c8f157be05R231
> 
> 99% sure that this fixes it by forcing cache busting when the hash of the file changes. At least I have not seen it since I turned that on

Oh -- it's a one off? In that case, meh.

bhearsum (approved):

PR #1751, opened 2025-08-05
Fix routes that require authentication

Eijebong:
So this actually never worked, not even before the react updates as both `new` and `newxpi` weren't marked  as requiring authentication and they would just silently throw errors instead. With this, if you're unauthenticated it'll at least redirect you to the index which should show that you're not authed.

Also we can see how great of a language this is (/s) because `Redirect` doesn't even exist anymore and wasn't imported, neither was `Suspense`, and the `render` property doesn't exist on `Route` anymore... And yet, nothing complained about any part of it...


Eijebong:
This introduces a very annoying white flash when the page first loads (assuming because of that `<Suspense>`)

PR #1766, opened 2025-08-20
config: bump CURRENT_ESR to 140 (bug 1984112)

jcristau:


Eijebong (approved):
Approving early so you're not blocked when we actually need to merge this

PR #1771, opened 2025-08-22
Bug 1983817 - Move the CORS middleware before exceptions

Eijebong:
The main issue with errors not being displayed is because those came from `flask.abort` which made routes raise `HttpException`s, which would be intercepted by the ASGI Exception middleware and not forwarded to the CORS one leading to those responses having the right info in them but not the right headers so the frontend wouldn't be able to read them, leading to network errors.

BEFORE_EXCEPTION is the recommended position in the documentation https://connexion.readthedocs.io/en/latest/cookbook.html#cors

### DIFF
--- a/api/src/backend_common/cors.py
+++ b/api/src/backend_common/cors.py
@@ -13,7 +13,7 @@ def init_app(app: connexion.App):
     origins = origins.split(" ")
     app.add_middleware(
         CORSMiddleware,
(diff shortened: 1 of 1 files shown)


jcristau (approved):
We should probably have tests for this, but given their absence for most of the admin API, not going to block on that.

PR #1773, opened 2025-08-25
Rename shipit's development environment to local and remove mentions of staging

Eijebong:
This is something that has always confused me because we had dev/development and I never know which one is localdev and which one is stage. Renamed it to local then found the TODO about it so I also removed mentions of staging while I was there.


ahal (approved):
Thanks, much better

PR #1785, opened 2025-09-02
Replace is-absolute-url with pure javascript

Eijebong:
https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static this does the exact same thing when not passed a second argument


jcristau (approved):
To be fair it looks like canParse is relatively recent?

PR #1793, opened 2025-09-19
api: order result from /xpi/releases

jcristau:
This fixes the recentxpi page to actually show the most recently shipped extensions.


hneiva (approved):
Thank you for noticing this 😄

PR #1796, opened 2025-10-08
Bug 1992745 - Improve cancellation of task groups resulting from an action task

Eijebong:
This fixes a race condition where cancelling a release while an action task for a phase was still running resulted in the resulting task group to not get cancelled at all.


Eijebong:
Scope changes for fxci https://github.com/mozilla-releng/fxci-config/pull/579

PR #1946, opened 2026-04-16
Fix Taskcluster secret name and XPI_LAX_SIGN_OFF docs

JohanLorenzo:
[Bug 2025695](https://bugzilla.mozilla.org/show_bug.cgi?id=2025695)

I wanted to run an entire staging release and I had to disturb another release engineer. I don't believe we need 2 people for a staging release.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,12 +68,13 @@ Setting up for the Public API:
     export FLASK_APP="shipit_api.public.flask:flask_app"
     uv run flask shell
 
-To provide all required sign offs on staging xpi releases
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+To provide all required sign offs on xpi releases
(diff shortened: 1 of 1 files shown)


Eijebong:
Isn't this what `XPI_LAX_SIGN_OFF=true` is about? Might be worth setting that to true on staging instead of inventing a second way of doing the same thing

JohanLorenzo:
Oh, thanks for catching this, @Eijebong! I wasn't aware of this env var. I agree, it totally makes sense. I repurposed this PR to update the documentation. I'm planning to add `XPI_LAX_SIGN_OFF=true` to both dev and staging. How does that sound?

Eijebong (approved):

Eijebong:
> I'm planning to add XPI_LAX_SIGN_OFF=true to both dev and staging. How does that sound?

SGTM

JohanLorenzo:
Done!

PR #1959, opened 2026-05-06
frontend: use CheckCircleOutlinedIcon instead of deprecated CheckCirc…

jcristau:
…leOutlineIcon

Per https://mui.com/material-ui/migration/upgrade-to-v9/#material-icons they're duplicates and the "Outline" version is removed in v9.


jcristau:
Should help unblock #1956.

Eijebong (approved):

PR #1971, opened 2026-05-22
feat: fix `shipit_import` command and make it more useful

bhearsum:
I was trying to test the rebuild product details functionality locally and found that was impractical (if not impossible) to get the right local data to do so. It seems we had the ability to import data from an existing instance once upon a time, but it long since broke. This pull request fixes it and adds the minimal enhancements needed to get a local database in shape enough to rebuild product details (import some or all releases; import nightly version information).

With this patch, `shipit_rebuild_product_details` is able to regenerate product details locally after `shipit_import` is run.

This is probably best reviewed commit-by-commit.

### DIFF
--- a/api/src/shipit_api/admin/cli.py
+++ b/api/src/shipit_api/admin/cli.py
@@ -22,7 +22,7 @@ from backend_common.log import configure_logging
 from shipit_api.admin.flask import flask_app
 from shipit_api.admin.product_details import rebuild
 from shipit_api.common.config import BREAKPOINT_VERSION
-from shipit_api.common.models import Release
+from shipit_api.common.models import Release, Version
 
 
 def coroutine(f):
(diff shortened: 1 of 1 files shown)


api/src/shipit_api/admin/cli.py
```diff
@@ -114,34 +115,42 @@ def get_taskcluster_headers(request_url, method, content, taskcluster_client_id,
...
+            req = requests.get(f"{api_from}/releases?status=shipped")
+            req.raise_for_status()
+            releases = req.json()
+            # Pull the most recent N releases when limiting; they're more important
+            # for testing in most cases.
+            releases.sort(key=lambda r: datetime.fromisoformat(r["created"]), reverse=True)
+            if limit_releases == -1:
+                limit_releases = releases.len()
+
+            for release in releases[0:limit_releases]:
```
ahal: nit: `[:limit_releases]` is a bit more idiomatic

ahal (approved):
Thanks for making this work again!

PR #1972, opened 2026-05-22
feat: add support for tracking nightly release metadata

bhearsum:
As part of https://bugzilla.mozilla.org/show_bug.cgi?id=2031045 I need a way to cheaply find the first version a locale shipped with. The solution I'm aiming at is making this data available in product-details. We already gather enough data as part of rebuilding it to make this data available for releases. We don't have it for nightlies. Let's put this data into Ship It, similar to how we're [already tracking current nightly version information](https://github.com/mozilla-releng/shipit/blob/fb4469b28650c6dbcb78e3758c181e5b82ea089a/api/src/shipit_api/common/models.py#L323). After this is landed we'll need to do a backfill of Nightly data by hand, and update it in automation (similar to how we [add current nightly version information](https://searchfox.org/firefox-main/source/taskcluster/kinds/release-update-product-channel-version/kind.yml)) before we can start putting this data in product details.

In theory, we could gather it by looking through VCS data or scraping archive -- in practice there's far too many nightlies for this to be practical (1,000s, if not 10,000s).

See also: https://bugzilla.mozilla.org/show_bug.cgi?id=2042596

### DIFF
--- a/api/src/backend_common/api.yml
+++ b/api/src/backend_common/api.yml
@@ -11,8 +11,101 @@ info:
     name: MPL 2.0
     url: https://www.mozilla.org/MPL/2.0/
   version: 1.0.0
+paths:
+  /nightly-release:
+    get:
+      summary: List nightly releases
+      operationId: shipit_api.public.api.list_nightly_releases
+      parameters:
+        - name: product
+          in: query
+          required: true
+          schema:
+            type: string
+        - name: channel
+          in: query
+          required: true
+          schema:
+            type: string
+        - name: version
+          in: query
+          schema:
+            type: string
+        - name: buildid
+          in: query
+          schema:
+            type: string
+        - name: after_buildid
+          in: query
+          description: >
+              Only return releases with buildid greater than this value. Feeding
+              the highest buildid returned by a query using this argument into a
+              new request as `after_buildid` will allow paging through entries
+              in an ascending fashion.
+          schema:
+            type: string
+        - name: before_buildid
+          in: query
+          description: >
+              Only return releases with buildid less than this value. Feeding
+              the lowest buildid returned by a query using this argument into a
+              new request as `before_buildid` will allow paging through entries
+              in a descending fashion.
+          schema:
+            type: string
+        - name: limit
+          in: query
+          description: Maximum number of releases to return.
+          schema:
+            type: integer
+            minimum: 1
+            maximum: 500
+            default: 500
+        - name: order
+          in: query
+          schema:
+            type: string
+            enum:
+              - desc
+              - asc
+            default: desc
+      responses:
+        "200":
+          description: A list of nightly releases
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/NightlyRelease'
 components:
   schemas:
+    NightlyRelease:
+      type: object
+      required:
+        - product
+        - channel
+        - version
+        - buildid
+        - locales
+      properties:
+        product:
+          type: string
+          example: firefox
+        channel:
+          type: string
+          example: nightly
+        version:
+          type: string
+          example: 140.0a1
+        buildid:
+          type: string
+          example: "20260522093015"
+        locales:
+          type: array
+          items:
+            type: string
+            example: zh-TW
     ProductInput:
       # ProductInput handles the products we currently support on shipit
       type: string
--- a/api/src/shipit_api/admin/api.yml
+++ b/api/src/shipit_api/admin/api.yml
@@ -825,6 +825,28 @@ paths:
         "409":
           description: Already submitted
           content: {}
+  /nightly-release:
+    # get handler is defined in backend_common
+    post:
+      summary: Add a new nightly release
+      operationId: shipit_api.admin.api.add_nightly_release
+      requestBody:
+        description: Nightly release object
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/NightlyRelease'
+        required: true
+      responses:
+        "201":
+          description: Nightly release added
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/NightlyRelease'
+        "400":
+          description: Invalid input
+          content: {}
   /versions/{product}/{channel}:
     get:
       summary: Get the current version for the given product channel
--- a/api/src/shipit_api/common/models.py
+++ b/api/src/shipit_api/common/models.py
@@ -327,3 +327,24 @@ class Version(db.Model):
     product_channel = sa.Column(sa.String, nullable=False)
     current_version = sa.Column(sa.String, nullable=False)
     __table_args__ = (sa.UniqueConstraint("product_name", "product_channel", name="_product_name_channel_uc"),)
+
+
+class NightlyRelease(db.Model):
+    __tablename__ = "shipit_api_nightly_releases"
+    id = sa.Column(sa.Integer, primary_key=True)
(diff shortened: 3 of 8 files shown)


api/src/shipit_api/admin/api.yml
```diff
@@ -1225,6 +1275,52 @@ components:
             - releases/mozilla-esr68
             - try
             - ''
+    NightlyReleaseInput:
```
jcristau: Do we need both the NightlyRelease and NightlyReleaseInput here?  They look identical.

jcristau: (as a side note we might want to move some definitions to backend_common/api.yml instead of duplicating them across public and admin)

api/src/shipit_api/public/api.py
```diff
@@ -94,3 +94,16 @@ def get_disabled_products():
     for row in session.query(DisabledProduct).all():
         ret[row.product].append(row.branch)
     return ret
+
+
+def list_nightly_releases(product=None, channel=None, version=None, buildid=None):
```
jcristau: Should e.g. the product be required?  We might want to think about ordering and pagination as well, if we're going to have more than a handful of these.

api/src/shipit_api/admin/api.yml
bhearsum: Yeah, these are good calls.

api/src/shipit_api/public/api.py
bhearsum: Good calls for sure on ordering and pagination...this is bound to be a 502 at some point if left as is.

On `product` being required or not, I could go either way. It seems unnecessary if we have pagination with an upper limit on the number of results returned?

jcristau: Yeah, it's fine, I just struggle to think of a use case where we wouldn't know the product.

bhearsum: Fair point.

I ended up implementing this (in part because it allowed me to avoid implementing a sort on `version`, which is non-trivial).

api/src/shipit_api/common/models.py
```diff
@@ -327,3 +327,24 @@ class Version(db.Model):
     product_channel = sa.Column(sa.String, nullable=False)
     current_version = sa.Column(sa.String, nullable=False)
     __table_args__ = (sa.UniqueConstraint("product_name", "product_channel", name="_product_name_channel_uc"),)
+
+
+class NightlyRelease(db.Model):
+    __tablename__ = "shipit_api_nightly_releases"
+    id = sa.Column(sa.Integer, primary_key=True)
```
jcristau, line 334: nit: it's harmless, but not sure we need this `id` for anything?

api/src/shipit_api/public/api.py
```diff
@@ -94,3 +94,24 @@ def get_disabled_products():
...
         ret[row.product].append(row.branch)
     return ret
+
+
+def list_nightly_releases(product, channel, version=None, buildid=None, page=1, per_page=500, order="desc"):
+    # unstable pagination is used here because:
+    # 1) the data does not change often
+    # 2) there are no known use cases that require stable pagination (the only
+    #    use case that requires all nightly release rows is rebuilding product
+    #    details, and that is done with direct database access)
```
jcristau: Not sure I buy this.  We could avoid using offset here and use a buildid as the criterion for which results to return, i.e. replace the `page=1` param with a `maxbuildid=None`, and if it's set, filter the query with `NightlyRelease.buildid<maxbuildid` (and same with a min).

bhearsum: While I think unstable pagination is good enough for the known use cases, min+max buildid seems like a better idea that's roughly the same amount of effort. Let's do it.

api/src/shipit_api/common/models.py
bhearsum: To be honest, I was mostly following the style here: a separate id as the PK like with the other tables.

api/src/shipit_api/public/api.py
```diff
@@ -94,3 +94,23 @@ def get_disabled_products():
...
+def list_nightly_releases(product, channel, version=None, buildid=None, minbuildid=None, maxbuildid=None, limit=500, order="desc"):
+    query = NightlyRelease.query.filter(NightlyRelease.product == product).filter(NightlyRelease.channel == channel)
+    if version:
+        query = query.filter(NightlyRelease.version == version)
+    if buildid:
+        query = query.filter(NightlyRelease.buildid == buildid)
+    if minbuildid:
+        query = query.filter(NightlyRelease.buildid >= minbuildid)
+    if maxbuildid:
+        query = query.filter(NightlyRelease.buildid <= maxbuildid)
```
jcristau: Not sure about the `=`.  I figured the way you'd use this (assuming `desc` order) is get the first 500, extract the min buildid from those, and pass that as the maxbuildid param in the next request, in which case you want `<` not `<=`?

jcristau (approved):
Thanks.

api/src/shipit_api/public/api.py
bhearsum: Yeah, that would be smarter. Sorry...should've thought about that before posting for review again.

bhearsum: Updated to make this change, and also renamed the query args to make it more obvious that the values are exclusive now.

api/src/backend_common/api.yml
```diff
@@ -11,8 +11,99 @@ info:
...
+            type: string
+        - name: buildid
+          in: query
+          schema:
+            type: string
+        - name: after_buildid
+          in: query
+          description: >
+              Only return releases with buildid greater than this value. Feeding
+              the highest buildid returned is `after_buildid` into a new request will
```
jcristau: `is` doesn't seem like the right word here?

api/src/backend_common/api.yml
```diff
@@ -11,8 +11,99 @@ info:
...
+              allow paging through entries in an ascending fashion.
+          schema:
+            type: string
+        - name: before_buildid
+          in: query
+          description: >
+              Only return releases with buildid less than this value. Feeding
+              the lowest buildid returned is `before_buildid` into a new request will
```
jcristau: same

api/src/backend_common/api.yml
```diff
@@ -11,8 +11,99 @@ info:
...
+              allow paging through entries in an descending fashion.
```
jcristau: s/an/a/?

api/src/backend_common/api.yml
```diff
@@ -11,8 +11,99 @@ info:
     name: MPL 2.0
     url: https://www.mozilla.org/MPL/2.0/
   version: 1.0.0
+paths:
+  /nightly-release:
+    get:
+      summary: List nightly releases
+      operationId: shipit_api.public.api.list_nightly_releases
```
jcristau, line 18: Is referencing shipit_api.public from backend_common OK?  (I guess so, there's references to it from the admin api.yml already, it just surprised me a bit)

jcristau (approved):

api/src/backend_common/api.yml
bhearsum: There's no technical reason...although I can see why it's surprising and maybe confusing. I guess the alternatives here are duplication of the GET endpoint or moving the handler into `backend_common`. The latter would be worse IMO. The former might be okay at the risk of future diversion. (I'm inclined to just leave this as is for now.)

PR #1987, opened 2026-06-11
feat: build firefox_history_locales.json for the nightly channel

bhearsum:
This adds initial support for the new `firefox_history_locales.json` file described in https://bugzilla.mozilla.org/show_bug.cgi?id=2042596.

It's focused on just the Nightly channel initially; I plan to follow-up in fairly short order with support for other channels.

In order for this to work correctly in production I'll need to first backfill with metadata from each Nightly that added locales, and a recent Nightly. With that done, we'll end up with [this firefox_locales.json](https://gist.github.com/bhearsum/cff2f1bef5cd0d81b90b7e92a1f72c92). (I have that metadata and a script prepared locally to do this; I can share it if it's useful.)

We should also wait until [automatic nightly submission](https://phabricator.services.mozilla.com/D304523) is landed to merge this to ensure it updates correctly the next time a locale is added or removed.

### DIFF
--- a/api/src/shipit_api/admin/product_details.py
+++ b/api/src/shipit_api/admin/product_details.py
@@ -8,6 +8,7 @@ import collections
 import functools
 import hashlib
 import io
+import itertools
 import json
 import logging
 import os
@@ -37,6 +38,8 @@ from shipit_api.admin.release import parse_version
 from shipit_api.common.product import Product, ProductCategory
 
 logger = logging.getLogger(__name__)
+
+NIGHTLY_RELEASES_BATCH_SIZE = 1000
 
 File = str
 ReleaseDetails = TypedDict(
@@ -305,12 +308,16 @@ def get_releases_from_db(db_session: sqlalchemy.orm.Session, breakpoint_version:
     return query.all()
 
 
-def get_nightly_releases_from_db(db_session: sqlalchemy.orm.Session, product: str, channel: str) -> typing.List[shipit_api.common.models.NightlyRelease]:
+def get_nightly_releases_from_db(db_session: sqlalchemy.orm.Session, product: str, channel: str) -> typing.Iterator[shipit_api.common.models.NightlyRelease]:
     NightlyRelease = shipit_api.common.models.NightlyRelease
     query = db_session.query(NightlyRelease)
     query = query.filter(NightlyRelease.product == product)
     query = query.filter(NightlyRelease.channel == channel)
-    return query.all()
+    # batching and ordering avoids loading all objects (which could be 10,000+) at once,
+    # and allows the callers to bail early if it finds all necessary information before
+    # iterating over all rows
+    query = query.order_by(NightlyRelease.buildid.desc())
+    return query.yield_per(NIGHTLY_RELEASES_BATCH_SIZE)
 
 
 def get_product_channel_version(db_session: sqlalchemy.orm.Session, product: str, channel: str):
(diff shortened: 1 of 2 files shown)


bhearsum:
I should also add: https://bugzilla.mozilla.org/show_bug.cgi?id=2047022 is looking at getting `en-US` added to `all-locales`, which would have the side effect of having it [submitted as part of the locale list](https://phabricator.services.mozilla.com/D304523) (which seems like the correct thing to do setting aside product details IMO). _That_ will have the side effect of including `en-US` in `firefox_locales.json`. I'm a bit torn on whether or not we should include `en-US` in `firefox_locales.json`. If we do, we'll probably want to make sure the right metadata is in place to have its first release listed as 0.8 (or maybe 1.0?).

api/src/shipit_api/admin/product_details.py
```diff
@@ -1222,6 +1318,7 @@ async def rebuild(
             breakpoint_version, Product.FIREFOX, combined_releases, combined_l10n, old_product_details, firefox_nightly_version, thunderbird_nightly_version
         ),
         "firefox_versions.json": await get_firefox_versions(releases, firefox_nightly_version),
+        "firefox_locales.json": get_firefox_locales(releases, firefox_nightly_releases),
```
ahal: Would it be better to call this `firefox_locale_history.json` or similar? Or do you anticipate other types of metadata being added in the future

api/src/shipit_api/admin/product_details.py
```diff
@@ -1043,6 +1066,75 @@ def get_thunderbird_beta_builds() -> typing.Dict:
...
 
+def get_firefox_nightly_locales(nightly_releases: typing.List[shipit_api.common.models.NightlyRelease]) -> FirefoxLocales:
+    """Generate the first version each locale has been continuously available on
+    for nightly releases.
+    """
+    builds = sorted(nightly_releases, key=lambda nightly: nightly.buildid, reverse=True)
+    # locales not present in the latest build are ignored completely, as they do
+    # not meet the definition of "continuously available"; make these easily
+    # available
+    latest_build_locales = builds[0].locales
```
ahal: nit: maybe worth handling the case where there are no builds in the DB in case we're ever deploying a fresh instance

ahal (approved):
Thanks, lgtm!

api/src/shipit_api/admin/product_details.py
bhearsum: I'm not terribly fussed about the name either way, and don't have plans to add other metadata. I think this is a good suggestion, although maybe `firefox_history_locales.json` would be better (to line up with files such as [firefox_history_major_releases.json](https://product-details.mozilla.org/1.0/firefox_history_major_releases.json))?

api/src/shipit_api/admin/product_details.py
bhearsum: Agreed, it's worth being safer here.

jcristau:
Any thoughts on what happens / should happen for locales that are discontinued?

bhearsum:
> Any thoughts on what happens / should happen for locales that are discontinued?

To be honest, I wasn't planning to do anything about them just yet. I don't think we need them to support the use case I'm adding this for (update integrity tests) because such locales won't be present in `l10n-changesets.json`, and won't be tested at all.

bhearsum:
(The easy/stupid thing to do would be to add `last_release`, but this might get weird for locales that have been added and removed multiple times. I don't know if there's a way to do better there short of querying all metadata.)

api/src/shipit_api/admin/product_details.py
```diff
@@ -290,6 +305,14 @@ def get_releases_from_db(db_session: sqlalchemy.orm.Session, breakpoint_version:
     return query.all()
 
 
+def get_nightly_releases_from_db(db_session: sqlalchemy.orm.Session, product: str, channel: str) -> typing.List[shipit_api.common.models.NightlyRelease]:
+    NightlyRelease = shipit_api.common.models.NightlyRelease
+    query = db_session.query(NightlyRelease)
+    query = query.filter(NightlyRelease.product == product)
+    query = query.filter(NightlyRelease.channel == channel)
+    return query.all()
```
jcristau: Can we avoid grabbing all nightlies?  This seems unlikely to scale well.

jcristau (changes requested):

api/src/shipit_api/admin/product_details.py
bhearsum: Yes, absolutely.

bhearsum: Added this in a subsequent commit.

jcristau (approved):

PR #1996, opened 2026-06-16
Bug 2047438 - Add ipp-activator to the list of system add-ons

bakulf:


bhearsum (approved):
We have a scheduled prod deploy on Thursday. If you need this in sooner than that please let me know.

PR #1999, opened 2026-06-22
feat: add support for importing nightly releases

bhearsum:
...and also add some tests for the `shipit-import` command.

### DIFF
--- a/api/src/shipit_api/admin/cli.py
+++ b/api/src/shipit_api/admin/cli.py
@@ -23,7 +23,7 @@ from backend_common.log import configure_logging
 from shipit_api.admin.flask import flask_app
 from shipit_api.admin.product_details import rebuild
 from shipit_api.common.config import BREAKPOINT_VERSION
-from shipit_api.common.models import Release, Version
+from shipit_api.common.models import NightlyRelease, Release, Version
 
 
 def coroutine(f):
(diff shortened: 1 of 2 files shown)


api/src/shipit_api/admin/cli.py
```diff
@@ -156,6 +155,57 @@ def shipit_import(api_from, limit_releases):
...
+                    req.raise_for_status()
+                    nightlies = req.json()
+                    if not nightlies:
+                        break
+
+                    for nightly in nightlies:
+                        if limit_releases != -1 and imported >= limit_releases:
+                            break
+                        before_buildid = nightly["buildid"]
+                        imported += 1
```
jcristau, line 181: should increment this be after the release "already exists" check?

jcristau (approved):
Wondering if this could use some batching instead of doing a ton of db round trips...

api/src/shipit_api/admin/cli.py
bhearsum: Yes!

bhearsum:
> Wondering if this could use some batching instead of doing a ton of db round trips...

Yes...but that sounds like a problem for the future. (It works fine as is even with no limit...and I'm putting in minimal effort on this right now.)

api/src/shipit_api/admin/cli.py
bhearsum: https://github.com/mozilla-releng/shipit/pull/2002 for this

PR #2000, opened 2026-06-22
fix: don't allow empty locales to be submitted in nightly releases

bhearsum:
https://bugzilla.mozilla.org/show_bug.cgi?id=2049378 discovered that this was allowed; let's reject them in the future.

### DIFF
--- a/api/src/shipit_api/admin/api.py
+++ b/api/src/shipit_api/admin/api.py
@@ -430,6 +430,9 @@ def add_nightly_release(body):
         user_permissions = ", ".join(current_user.get_permissions())
         abort(401, f"required permission: {required_permission}, user permissions: {user_permissions}")
 
+    if any(len(l) == 0 for l in body["locales"]):
+        abort(400, "entry in locales contains empty string")
+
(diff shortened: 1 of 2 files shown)


bhearsum:
Leaving as a draft until https://bugzilla.mozilla.org/show_bug.cgi?id=2049378 is on main, otherwise we'll fail those tasks.

api/src/shipit_api/admin/api.py
```diff
@@ -430,6 +430,9 @@ def add_nightly_release(body):
         user_permissions = ", ".join(current_user.get_permissions())
         abort(401, f"required permission: {required_permission}, user permissions: {user_permissions}")
 
+    if any([len(l) == 0 for l in body["locales"]]):
```
jcristau:
```suggestion
    if any(len(l) == 0 for l in body["locales"]):
```

jcristau (approved):
just one small nit

bhearsum:
Gecko change is on main; this is ready to go now.

PR #2001, opened 2026-06-23
Bug 2046486 - add support for esr153

jcristau:


bhearsum (changes requested):
It looks like https://github.com/mozilla-releng/shipit/blob/main/api/trust-domains.yml needs to be updated as well.

jcristau:
Updated, thanks!

bhearsum (approved):

PR #2010, opened 2026-07-10
fix: stop using management api auth0 audience

bhearsum:
This looks like a longstanding mistake. Recently, logins started failing with errors about auth.mozilla.auth0.mozilla.com/api/v2 -- which is the management API. The IAM team confirmed that they recently locked this down more, which explains why this came up all of a sudden.

As far as I know, we don't make any use of the management API in Ship It (this is the API that's used to make changes to clients, etc.), so this should be safe.

I've tested this in Ship It Dev, where I was able to login again and cancel a release.

### DIFF
--- a/frontend/src/App.jsx
+++ b/frontend/src/App.jsx
@@ -27,7 +27,6 @@ const App = () => {
           clientId={config.AUTH0.clientID}
           authorizationParams={{
             redirect_uri: config.AUTH0.redirectUri,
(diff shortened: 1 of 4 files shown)


bheesham (approved):

bhearsum:
FYI: I am merging this without releng review because I can't find a reviewer this late, and it needs to be fixed. I'd appreciate a post-facto review.

PR #2009, opened 2026-07-08
bug 2042596: add non-nightly channels to firefox_history_locales.json

bhearsum:
This is mostly straightforward with the exception how DevEdition/aurora is handled. As noted in the comments, I'm largely pretending that it didn't exist prior to 54.0b11. This means that the information for this channel is not _strictly_ accurate, but it allows me to sidestep a lot of complexity that is unlikely to benefit any real use case. I'm open to disagreement on this point.

https://gist.github.com/bhearsum/2ad4587c3b638b8fa6d666e334e40a25 contains a locally generated file from a relatively recent import from ship it production. https://github.com/mozilla-releng/product-details/commit/9d913e824e0df9013c612dc997f804d41dace2b2 is a commit made by ship it dev with this patch applied.

PR #2011, opened 2026-07-14
bug 2042596: add non-nightly channels to firefox_history_locales.json

bhearsum:
This is mostly straightforward with the exception how DevEdition/aurora is handled. As noted in the comments, I'm largely pretending that it didn't exist prior to 54.0b11. This means that the information for this channel is not _strictly_ accurate, but it allows me to sidestep a lot of complexity that is unlikely to benefit any real use case. I'm open to disagreement on this point.

https://gist.github.com/bhearsum/3ef0f84fd7f95f0cfb1a96c55b096dc3 contains a locally generated file from a relatively recent import from ship it production. https://github.com/mozilla-releng/product-details/blob/923f0c8fd053ad595218e24f2782364c18b7200e/public/1.0/firefox_history_locales.json is the latest version generated in dev. (Note that outside of Nightly, most locales don't exist because staging releases are typically run with a reduced locale set.)

This is a continuation of the work from https://github.com/mozilla-releng/shipit/pull/2009.

### DIFF
--- a/api/src/shipit_api/admin/product_details.py
+++ b/api/src/shipit_api/admin/product_details.py
@@ -39,6 +39,7 @@ from shipit_api.common.product import Product, ProductCategory
 
 logger = logging.getLogger(__name__)
 
+AURORA_FIRST_VERSION = "54.0b11"
 NIGHTLY_RELEASES_BATCH_SIZE = 1000
 
 File = str
@@ -111,7 +112,7 @@ ThunderbirdVersions = TypedDict(
         "THUNDERBIRD_ESR_NEXT": str,
     },
 )
-FirstRelease = TypedDict("FirstRelease", {"version": str, "buildid": str})
+FirstRelease = TypedDict("FirstRelease", {"version": typing.Required[str], "buildid": str, "build_number": int}, total=False)
 LocaleFirstReleases = TypedDict("LocaleFirstReleases", {"first_release": typing.Dict[str, FirstRelease]})
 FirefoxLocales = typing.Dict[str, LocaleFirstReleases]
 IndexListing = str
@@ -173,6 +174,24 @@ def dt_to_ymd(d: datetime) -> str:
 
 def from_ymd_format(s: str) -> datetime:
     return from_format(s, YMD_DATE_FORMAT)
+
+
+def get_channel(version, product_category):
+    # some ESRs are marked with the "major" category; the only way to
+    # reliably detect them is to look at the version number. this must
+    # be done first to avoid eg: 140.0esr (a "major" release) returning
+    # "release" as the channel
+
+    if version.endswith("esr"):
+        return "esr"
+
+    match product_category:
+        case ProductCategory.DEVELOPMENT.value:
+            return "beta"
+        case ProductCategory.MAJOR.value | ProductCategory.STABILITY.value:
+            return "release"
+
+    raise Exception(f"Can't detect channel for version: {version}, category: {product_category}!")
 
 
 def create_index_listing_html(folder: pathlib.Path, items: typing.Set[pathlib.Path]) -> str:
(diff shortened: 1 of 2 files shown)


api/src/shipit_api/admin/product_details.py
```diff
@@ -1073,74 +1092,154 @@ def get_thunderbird_beta_builds() -> typing.Dict:
...
         # once every locale of interest has had its oldest continuous release
-        # identified, no older build can change the result; stop fetching
+        # identified, no older release can change the result; stop fetching
         if done == needed_locales:
             break
 
+    return last_releases
+
+
+def get_firefox_nightly_locales(nightly_releases: typing.Iterable[shipit_api.common.models.NightlyRelease]) -> FirefoxLocales:
```
hneiva, line 1142: nit: importing some of these directly would improve readability across the file

(maybe this belongs in a different PR)

api/src/shipit_api/admin/product_details.py
```diff
@@ -1073,74 +1092,154 @@ def get_thunderbird_beta_builds() -> typing.Dict:
...
-    return {}
+    channels = {}
+    # Combine release information with locales for each release
+    for product_with_version, details in firefox_releases["releases"].items():
+        version = product_with_version.split("-", 1)[1]
+        build_number = details["build_number"]
+        channel = get_channel(version, details["category"])
+
+        if channel is None:
+            continue
```
hneiva: `channel` can never be `None`

api/src/shipit_api/admin/product_details.py
```diff
@@ -1073,74 +1092,154 @@ def get_thunderbird_beta_builds() -> typing.Dict:
...
+        else:
+            data["first_release"]["aurora"] = beta.copy()
+
+    return result
 
 
 def get_firefox_locales(
-    releases: typing.List[shipit_api.common.models.Release], nightly_releases: typing.List[shipit_api.common.models.NightlyRelease]
+    firefox_releases: Releases,
+    releases_l10n: ReleaseL10n,
```
hneiva: Looks like this should be `releases_l10n: Dict[Release, ReleaseL10ns]`.

Also worth adding the annotation to `get_firefox_release_locales`

api/src/shipit_api/admin/product_details.py
```diff
@@ -1073,74 +1092,154 @@ def get_thunderbird_beta_builds() -> typing.Dict:
...
+
+        if locales is None:
+            # We don't have consistently available l10n information for versions < 40
+            # Don't error out unless we're missing data for newer versions
+            if parse_version(Product.FIREFOX, version).major_number > 39:
+                raise Exception(f"Couldn't find l10n information for version: {version}")
+            else:
+                continue
+
+        release = {"version": version, "build_number": build_number}
```
hneiva: nit: This is reusing the same variable name as the loop over `releases_l10n`. Not a biggie, but might be confusing.

hneiva (approved):
A few minor comments, overall looks great.

api/src/shipit_api/admin/product_details.py
bhearsum: Yeah, I agree. I was just following existing style here. I'll consider a follow-up though.

api/src/shipit_api/admin/product_details.py
bhearsum: Yeah, fair point - I'll fix this.

api/src/shipit_api/admin/product_details.py
```diff
@@ -111,7 +112,7 @@
         "THUNDERBIRD_ESR_NEXT": str,
     },
 )
-FirstRelease = TypedDict("FirstRelease", {"version": str, "buildid": str})
+FirstRelease = TypedDict("FirstRelease", {"version": str, "buildid": str, "build_number": int}, total=False)
```
jcristau: nit: `"version": Required[str]`?

api/src/shipit_api/admin/product_details.py
```diff
@@ -1145,16 +1164,81 @@ def yield_releases() -> typing.Iterable[tuple[FirstRelease, list[str]]]:
...
+            for release, l10n_changesets in releases_l10n.items():
+                if release.product == details["product"] and release.version == version and release.build_number == details["build_number"]:
+                    locales = l10n_changesets.keys()
+
+        if locales is None:
+            # We don't have consistently available l10n information for versions < 40
+            # Don't error out unless we're missing data for newer versions
+            if parse_version(Product.FIREFOX, version).major_number > 39:
+                raise Exception(f"Couldn't find l10n information for version: {version}")
+            else:
```
jcristau: nit: could remove this `else`, the `if` branch doesn't return (or switch it around to `if version < 40: continue; raise`), to match the comment

jcristau (approved):

api/src/shipit_api/admin/product_details.py
bhearsum: Good call.

PR #2027, opened 2026-07-21
feat: support importing merge automation data

bhearsum:
I'm working on some enhancements to merge automation; being able to import this data from production would be helpful.

I didn't implement any limitations on how many can be important because:
1) These don't build up very quickly (in ~2 months we've had 19 rows added)
2) They're cheap to import
3) They're all returned at the same time anyways; limiting the imports only saves work on the client, not the server

### DIFF
--- a/api/src/shipit_api/admin/cli.py
+++ b/api/src/shipit_api/admin/cli.py
@@ -23,7 +23,7 @@ from backend_common.log import configure_logging
 from shipit_api.admin.flask import flask_app
 from shipit_api.admin.product_details import rebuild
 from shipit_api.common.config import BREAKPOINT_VERSION
-from shipit_api.common.models import NightlyRelease, Release, Version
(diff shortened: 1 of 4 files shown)


api/src/shipit_api/admin/cli.py
```diff
@@ -224,6 +224,40 @@ def shipit_import(api_from, limit_releases):
...
+                    task_id=merge["task_id"],
+                    version=merge["version"],
+                    commit_message=merge["commit_message"],
+                    commit_author=merge["commit_author"],
+                    repo=merge["repo"],
+                    pretty_name=merge["pretty_name"],
+                    project=merge["project"],
+                )
+                session.add(m)
+                session.commit()
```
jcristau: Could we batch them instead of committing one at a time?

jcristau (approved):

